PR #6, opened 2015-08-03
Constant time encoder

klauspost:
Add encoder that does not attempt to do any back-matches. This gives much more predictable performance, but at a big compression size cost. Performance is currently approximately ~70MB/s.

Use level = ConstantCompression (-2) to enable.


klauspost:
Speed is now 110-160MB/sec, benchmarks added at https://docs.google.com/spreadsheets/d/1nuNE2nPfuINCZJRMt6wFWhKpToF95I47XjSsc-1rbPQ/edit?usp=sharing

PR #8, opened 2015-09-30
add a fuzzer

aybabtme:
Ohai!  I'm running a fuzzer on the gzip code, haven't found any crasher so far (wow!), but I thought I'd contribute back the code, if you're interested!

I'll let you know if the fuzzer ever finds a crasher.


aybabtme:
Forgot to mention that most of the corpus, and the `Fuzz` func, comes from https://github.com/dvyukov/go-fuzz/tree/master/examples/gzip

klauspost:
I didn't really expect any input-based problems, since the standard library deflate has been heavily tested. The most critical 

Would it be possible to put it in the main directory, so it isn't included in the gzip package itself? 

That way people don't have to download all the go-fuzz tests if they only want to use the package.

klauspost:
Added as a separate repo: https://github.com/klauspost/compress-fuzz

PR #25, opened 2015-12-17
Add Go (BSD) License.

### LINKED ISSUE #24: Missing BSD license
Is looks like some of the files are licensed under BSD:

> // Copyright 2011 The Go Authors. All rights reserved.
> // Use of this source code is governed by a BSD-style
> // license that can be found in the LICENSE file.

BSD says the following:

> 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

You are not respecting this condition, as there is no BSD license in this repository.

Depending on your intentions, you may want to do the following:
 - relicense klauspost/compress under BSD
 - Include the full license on all headers
 - update the headers and mention a ```LICENSE.bsd``` file instead, and then include ```LICENSE.bsd``` at the root of the repository.

klauspost:
Fixes #24

### DIFF
--- a/GO_LICENSE
+++ b/GO_LICENSE
@@ -1,0 +1,27 @@
+Copyright (c) 2012 The Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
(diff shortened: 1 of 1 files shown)


klauspost:
I think i will just relicense instead (since it is going back to Go soon).

PR #27, opened 2015-12-20
Remove references of project gutenberg

### LINKED ISSUE #23: Get rid of the ebook
Hey,

Would you mind getting rid of the ebook? (```testdata/Mark.Twain-Tom.Sawyer.txt```)

I am packaging compress for Debian and I don't think this book is free software ;). Maybe replace it with gibberish?

aviau:
Hello, this closes #23 

This is needed because the Project Gutenberg License is not acceptable for Debian. I didn't read the whole license, but I can tell right away because of conditions like this:

>1.E.8. You may charge a reasonable fee for copies of or providing access to or distributing Project Gutenberg-tm electronic works provided that:
- You pay a royalty fee of 20%

That would mean that selling a Debian CD requires to pay a royalty to Project Gutenberg.

to get rid of the Project Gutenberg license, all we have to do is remove the references to Project Gutenberg:

>  If you strip the Project Gutenberg license and all references to Project Gutenberg from the text, you are left with a text unprotected by U.S. intellectual property law. You can do anything you want with that text in the United States and most of the rest of the world.

I'll open a similar PR to the Golang project.


klauspost:
Thanks!

PR #31, opened 2016-01-30
Rebalance compression levels.

klauspost:
This will rebalance the compression levels to have a more even distribution of compression levels. The general trend of this is to make compression faster, which means that a given compression level (except 1 and 9) will be faster.

The existing implementation favors compression, and from level 5 and up, very small compresssion improvements are added for very high computation cost.

Rough guidelines: Previous level 3 speed is now equal to level 6.

New default is level 5, which in most cases offers a quite good compression, without getting diminished returns from additional computation power.

See "rebalance" tabs for benchmarks of rebalanced levels:

https://docs.google.com/spreadsheets/d/1nuNE2nPfuINCZJRMt6wFWhKpToF95I47XjSsc-1rbPQ/edit?usp=sharing

### DIFF
--- a/flate/deflate.go
+++ b/flate/deflate.go
@@ -1094,7 +1094,7 @@ func (d *compressor) storeSnappy() {
 	if d.tokens.n == d.windowEnd {
 		d.err = d.writeStoredBlock(d.window[:d.windowEnd])
 		// If we removed less than 10 literals, huffman compress the block.
-	} else if d.tokens.n > d.windowEnd-10 {
+	} else if d.tokens.n > d.windowEnd-(d.windowEnd>>4) {
 		d.w.writeBlockHuff(false, d.window[:d.windowEnd])
 		d.err = d.w.err
 	} else {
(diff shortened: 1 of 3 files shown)


willglynn:
One thing I note about the original values: they seem to correspond to [the compression levels in zlib](https://github.com/madler/zlib/blob/master/deflate.c#L131-L143). Even the [`compress/flate` constants](https://golang.org/pkg/compress/flate/#pkg-constants) mirror the [zlib constants](https://github.com/madler/zlib/blob/v1.2.8/zlib.h#L186-L190). Someone who's used zlib before might assume that Go compression levels 0-9 are similar to zlib (I did!), and prior to this changeset, they'd be right.

I like the idea of having more speed/size tradeoffs available, but I'd also like a note in the docs saying that the compression levels work differently than zlib.

klauspost:
@willglynn You are right. It makes good sense to document why these values are different than zlib.

That said, in my tests zlib has always performed ~1 level worse than go deflate in terms of compression. Noting that you shouldn't expect settings to carry over 1:1 makes sense.

PR #46, opened 2016-03-13
Clarify that snappy is still gzip compatible

rasky:
I propose this file change to clarify that this modification still makes the output compatible with existing tools. It wasn't apparent to me at first read, and I double-checked with a couple of colleagues.


klauspost:
Thanks a lot for the clarification.

PR #51, opened 2016-04-14
Fix (!noasm AND !appengine) build tag.

nigeltao:
It was correct for the *_amd64.go files, just not the *_amd64.s files,
which used OR instead of AND.

See the "Build Constraints" section of https://golang.org/pkg/go/build/


klauspost:
Thanks!

PR #52, opened 2016-04-14
Remove DUPOK and NOPROF flags from asm functions.

nigeltao:
runtime/textflag.h defines:
NOPROF  1
DUPOK   2
NOSPLIT 4

The DUPOK bit was why the bug in
https://github.com/klauspost/compress/commit/6ea72a38a8cdbe25966059c547d425e580c624a7
wasn't picked up earlier.

The existing 7 is presumably a copy/paste of some standard library asm
code. It's a legacy idiom that really should be 5 or 4. See
https://groups.google.com/d/msg/golang-nuts/uJTYuruvnZM/6j4KXgkdcf0J for
some history.

Asm code in the standard library uses NOSPLIT instead of 4 these days.

### DIFF
--- a/flate/crc32_amd64.s
+++ b/flate/crc32_amd64.s
@@ -3,8 +3,10 @@
 
 // Copyright 2015, Klaus Post, see LICENSE for details.
 
+#include "textflag.h"
+
 // func crc32sse(a []byte) hash
-TEXT ·crc32sse(SB), 7, $0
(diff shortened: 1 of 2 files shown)


klauspost:
Would you mind changing it to `4`? I don't want to break backwards compatibility unless there is a compelling reason.

Thanks for the contribution.

nigeltao:
Done.

klauspost:
Thanks!

PR #60, opened 2016-08-20
WIP: Experiment on re-using dynamic encoding tables.

klauspost:
Gives a very nice 10-20% speedup on longer running encodings with quite small size cost, so worth pursuing.

Currently way to hacky and slows down single-block encoding.


klauspost:
Replaced by #105

PR #68, opened 2016-10-29
Bigger maxFlateBlockTokens

klauspost:
This increases performance at a very little size increase

Investigate if we should have this (and what it should be) for levels 5-7.


klauspost:
Included in #69

PR #100, opened 2019-04-23
zstd split sequence decode/execute

klauspost:
The basic idea is that we don't need the stream history to decode the sequences.

So we do history in two parts:

1) Decode as much as we can without any history.
2) Receive decoder history.
3) Decode sequences.
4) Send decoder history to next block. Next block goto 2)
5) Wait for stream history
6) Execute decoded sequences.
7) Send stream history to next block. Next block goto 6)

This should allow more concurrency for stream decoding.

We probably should keep the combined code for buffer decoding.

So this would require each decoder to have a channel on which it can receive the decoder history, if it is interested.

Estimated speed boost would be in the area of 1.5x

### DIFF
--- a/zstd/blockdec.go
+++ b/zstd/blockdec.go
@@ -653,13 +653,18 @@ func (b *blockDec) decodeCompressed(hist *history) error {
 		println("initializing sequences:", err)
 		return err
 	}
-
-	err = seqs.decode(nSeqs, br, hist.b)
+	s := b.sequenceBuf[:nSeqs]
+	err = seqs.decode(s, br)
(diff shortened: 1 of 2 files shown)


klauspost:
It does look like it is feasible to split sequence decoding and execution.

This is the execution time distribution on silencia corpus:

image

This could well be worth the synchronization cost.

klauspost:
Having execution always run on the same goroutine would also be a bonus. Not sure if that is feasible designwise.

klauspost:
Tracked in #477

PR #102, opened 2019-05-05
Zstd Encoder/Converter

klauspost:
Zstandard encoder.

Working, fine performance.

- [x] Determine a heuristic to reject potentially un-compressible blocks early to preserve offsets.
- [x] Simple encoder.
- [x] Fuzz testing.
- [x] Investigate reference encoder methods.
- [x] Make Encoder concurrently usable.
- [x] Add encoder options.
- [x] Add stream encoder.
- [x] Add async matching+block encoding.


zstd/encoder_options.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// WithEncoderCRC will add CRC value to output.
+// Output will be 4 bytes larger.
+func WithEncoderCRC(b bool) EOption {
+	return func(o *encoderOptions) error { o.crc = b; return nil }
+}
+
+// WithEncoderConcurrency will set the concurrency,
+// meaning the maximum number of decoders to run concurrently.
```
ribasushi, line 34: @klauspost is `decoders` in this helptext a typo, or does this encoder option indeed influence how many processes will run at decode-time?

klauspost: typo

PR #136, opened 2019-07-11
zstd: Fix decoder race.

klauspost:
... and add test for consecutive frames.


zstd/blockdec.go
```diff
@@ -145,18 +146,18 @@ func (b *blockDec) reset(br byteBuffer, windowSize uint64) error {
 	}
 
 	// Read block data.
-	if cap(b.data) < cSize {
+	if cap(b.dataStorage) < cSize {
 		if b.lowMem {
-			b.data = make([]byte, 0, cSize)
+			b.dataStorage = make([]byte, 0, cSize)
```
larry-cdn77, line 151: Is the allocation of `dataStorage` superfluous given that `readBig` doesn't actually use its second argument?

klauspost: @larry-cdn77 This does: https://github.com/klauspost/compress/blob/7ce6eae0ef5642fa10d811986686186ac85f4cf7/zstd/bytebuf.go#L99

larry-cdn77: Thanks for explaining. I can see how this works with stream decompression as opposed to regular decompression from a byte array. It seems that block decoder reset is the point at which bytes are extracted from the underlying I/O reader. The interface name `readerWrapper` gives a hint here. From this point onwards, stream decompression and regular byte-array decompression operate on the same kind of input (`blockDec.dst`).

The reason I looked into the use of `dataStorage` is that by not allocating it I get a 5% boost in my application throughput. My application obviously isn't stream based.

klauspost: @larry-cdn77

This should only be the case if you don't reuse the decoder, since this it kept between runs.

You can safely use a single decoder for multiple goroutines and after the initial warmup it should not impact performance.

larry-cdn77: Thanks once again. I am reusing the decoder, and indeed the performance boost is so small it may be down to a measurement error. I can now see how if `dataStorage` is allocated to a big enough size it will be reused without additional allocations.

PR #148, opened 2019-08-13
add go mod support

jsign:
Add `go mod` support.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/klauspost/compress
+
+go 1.12
+
+require (
+	github.com/google/go-cmp v0.3.1
+	github.com/klauspost/cpuid v1.2.1
+)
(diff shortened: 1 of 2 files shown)


jsign:
Is there any particular dependency version needed?
Having defined dependency versions (and vendoring) make performance benchmarks future-proof.

aaronbee:
I don't believe vendoring dependencies is appropriate here. go.mod will ensure all users of the library get the right dependencies without the need of the vendor directory.

jsign:
Well, that's is assuming that dependencies don't mess-up with mutating versions. This might be a reasonable assumption considering the particular small set of dependencies of this project; the worst case scenario will complain about hash-checks when `go mod download`.

Do you consider that might be enough and better removing `vendor` folder?

aaronbee:
Yes, you can assume dependencies don't mess up their versions. This is the reason for go.sum and the public notary (discussed [here](https://blog.golang.org/modules2019)) It would be a crummy dependency management system if you couldn't rely on a dependency version meaning the same thing to everyone.

Disclaimer: I'm just a fan of this particular repo.

jsign:
IMHO, vendoring isn't only about _version meaning_ but also availability. Considering that public modules version code is mirrored, vendoring might be redundant.

I think that's a fair point. :+1:

vtolstov:
Library dont need to have vendor dir, also go mod when you do go get don't download vendor folder
You need to remove it

klauspost:
I am not sure I see the need for explicit modules- it should work just fine.

#105 Should get rid of cpuid for now when it lands. And `go-cmp` should only be used by tests.

PR #149, opened 2019-08-18
Use unsafe src reading

klauspost:
About 4-5% faster :/


klauspost:
This would only work on little endian platforms. Not sure it is worth it.

PR #158, opened 2019-09-05
add WithWindowSize for configurable window size

ianwilkes:
Enabling smaller window sizes will save memory when working exclusively with smaller payloads. In addition, in some cases it offers faster compression with very similar results. Large window sizes are supported but are probably inadvisable in most situations.

### DIFF
--- a/zstd/encoder_options.go
+++ b/zstd/encoder_options.go
@@ -1,6 +1,7 @@
 package zstd
 
 import (
+	"errors"
 	"fmt"
 	"runtime"
 	"strings"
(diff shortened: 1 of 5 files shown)


klauspost:
Very nice!

I would like to have the default window size to depend on the compression level, which is why I didn't expose it and the relation isn't super clear.

I think except for promising a specific window size in the documentation, I don't see how adding this could be a problem.

zstd/encoder_options.go
```diff
@@ -64,6 +65,30 @@ func WithEncoderConcurrency(n int) EOption {
 	}
 }
 
+// WithWindowSize will set the maximum allowed back-reference distance.
+// The value must be a power of two between WindowSizeMin and WindowSizeMax.
+// A larger value will enable better compression but allocate more memory and,
+// for above-default values, take considerably longer.
+// By default this is set to 4MB.
```
klauspost:
```suggestion
// The default value is determined by the compression level.
```

klauspost:
Merged. Thanks for the contribution. It will be included in the next release!

PR #176, opened 2019-11-09
Add stateless gzip/deflate

klauspost:
`StatelessCompression` will do compression but without maintaining any state between Write calls.

There will be no memory kept between Write calls, but compression and speed will be suboptimal.

Because of this, the size of actual Write calls will affect output size.

In gzip, specify level `-3` to enable.

For direct deflate use, `NewStatelessWriter` and `StatelessDeflate` are available. See https://github.com/klauspost/compress/pull/176/files#diff-a5ae2d33d655db7f26bd8de2b9733ea5R43

A [`bufio.Writer`](https://golang.org/pkg/bufio/#Writer) can of course be used to control write sizes.


klauspost:
@FZambia re our talk.

vtolstov:
As i see this can be useful for some http wrappers that compress output?

klauspost:
@vtolstov Indeed. But you need to be able to control how writes are done. Writing one byte at the time = absolutely horrible.

PR #183, opened 2019-11-14
Add go vet to CI

klauspost:
Not all vets are run when tests run, so we add go vet explicitly.

Fix issues reported.

Thanks to @shawnps for fixing Go 1.12.


klauspost:
Hmmm... Go 1.12 seems to have an issue... FFS.

PR #198, opened 2019-12-26
Fix typo ration => ratio in README.md

jarifibrahim:
(I don't know why does the last line show up in the diff :man_shrugging: )


klauspost:
Thanks. It adds a newline which is 👍

PR #229, opened 2020-02-19
use previously defined size var when making dst byte slice

shawnps:


klauspost:
Definitely looks suboptimal. Thanks!

shawnps:
FWIW, I found this issue through `ineffassign`: 

https://goreportcard.com/report/github.com/klauspost/compress#ineffassign

There are more, but some seem like false positives. This one looks real:

https://github.com/klauspost/compress/blob/master/gzip/gzip_test.go#L214

But I wasn't sure how that file is maintained since it seems to be from the Go source. It appears fixed upstream:

https://github.com/golang/go/blob/master/src/compress/gzip/gzip_test.go#L216-L219

PR #230, opened 2020-02-19
run vet on all versions but go 1.12

shawnps:
@klauspost This is a bit weird, but I think running `go vet` is still useful even if you have to skip 1.12. Happy to throw this PR away if you think this is too much of a hack.


klauspost:
Do you mind if I include it in #183 which has the fixes I assume needs to be included as well?

shawnps:
@klauspost yes no problem, feel free to delete my commit, no need to do any cherry-picking if you don't feel like it.

shawnps:
@klauspost however, looks like it's not working:

https://travis-ci.org/klauspost/compress/jobs/652321614?utm_medium=notification&utm_source=github_status

Let me see if I can do some digging and find out why.

shawnps:
@klauspost through some trial and error I finally got this working if you want to bring it into your branch 👍 

The failing builds are due to legitimate vet errors as you mentioned. Builds on 1.12 skip vet and continue:

https://travis-ci.org/klauspost/compress/jobs/652357469?utm_medium=notification&utm_source=github_status

klauspost:
Thanks. Works perfectly in #183

Issue #248, opened 2020-03-26
zstd performance and Reader/Writer reuse


vmihailenco:
I have [a benchmark](https://gist.github.com/vmihailenco/1beb14cabe4ab597cefc3d6ab1361e3c) that produces following results:

```
BenchmarkCZstd-12      	     664	   1793805 ns/op	 125.62 MB/s	  522419 B/op	       9 allocs/op
BenchmarkGoZstd-12    	     374	   3247447 ns/op	  69.39 MB/s	 1646825 B/op	      12 allocs/op
```

That is roughly 2 times slower (and results can be worse on smaller payloads) and allocs are rather big (1646825 to encode & decode 225339 bytes?). I know that I can achieve better results by using `EncodeAll` and `DecodeAll`, but I would like to use Encoder/Decoder as wrappers over `bytes.Buffer`. So my question is - am I doing anything wrong here with Encoder/Decoder? I've disabled CRC for more fair comparison based on our previous discussion - I plan to use it in production.

klauspost:
Yes. There is some special handling when you decode a `*bytes.Buffer`, and it does indeed allocate more than it should. I will send a PR shortly.

As you note, the streaming en+decoders are not for payloads this small.

Also, if you want to eliminate more allocations you can copy data to a hash function or something so you avoid the `ioutil.ReadAll`.

klauspost:
Thanks for the detailed report 👍

vmihailenco:
Thanks - now it looks like this

```
BenchmarkCZstd-12      	     628	   1857834 ns/op	 121.29 MB/s	  522429 B/op	       9 allocs/op
BenchmarkGoZstd-12     	     423	   3165288 ns/op	  71.19 MB/s	  873224 B/op	      11 allocs/op
BenchmarkGoZstd2-12    	     523	   2133557 ns/op	 105.62 MB/s	  583841 B/op	       2 allocs/op
```

What still looks strange is that `EncodeAll/DecodeAll` version is faster and allocates less than using `Write` / `Read`. I guess there is a good reason for that, but from user POV I don't understand why there is such a big difference - after all `Encoder.Writer` can easily be a wrapper over `EncodeAll`.

klauspost:
>  after all Encoder.Writer can easily be a wrapper over EncodeAll.

No. With EncodeAll you always know how much you are compressing. Using a stream you don't know if the user will write more, so you need to buffer input and cannot start compressing until the stream is closed.

[...]

PR #251, opened 2020-03-27
zstd: If first block and 'final', encode direct.

klauspost:
If if writing header (ie first block) and it is the final block, use block compression instead of async.

Addition for #248

### DIFF
--- a/zstd/encoder.go
+++ b/zstd/encoder.go
@@ -172,6 +172,20 @@ func (e *Encoder) nextBlock(final bool) error {
 		return fmt.Errorf("block > maxStoreBlockSize")
 	}
 	if !s.headerWritten {
+		// If we have a single block encode, do a sync compression.
+		if final && len(s.filling) > 0 {
+			s.current = e.EncodeAll(s.filling, s.current[:0])
+			var n2 int
+			n2, s.err = s.w.Write(s.current)
+			if s.err != nil {
+				return s.err
+			}
+			s.nWritten += int64(n2)
+			s.current = s.current[:0]
+			s.filling = s.filling[:0]
+			return nil
+		}
+
 		var tmp [maxHeaderSize]byte
 		fh := frameHeader{
 			ContentSize:   0,
--- a/zstd/encoder.go
+++ b/zstd/encoder.go
@@ -39,17 +39,18 @@ type encoder interface {
 }
 
 type encoderState struct {
-	w             io.Writer
-	filling       []byte
-	current       []byte
-	previous      []byte
-	encoder       encoder
-	writing       *blockEnc
-	err           error
-	writeErr      error
-	nWritten      int64
-	headerWritten bool
-	eofWritten    bool
+	w                io.Writer
+	filling          []byte
+	current          []byte
+	previous         []byte
+	encoder          encoder
+	writing          *blockEnc
+	err              error
+	writeErr         error
+	nWritten         int64
+	headerWritten    bool
+	eofWritten       bool
+	fullFrameWritten bool
 
 	// This waitgroup indicates an encode is running.
 	wg sync.WaitGroup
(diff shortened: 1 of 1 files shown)


zstd/encoder.go
```diff
@@ -172,6 +172,20 @@ func (e *Encoder) nextBlock(final bool) error {
...
+			var n2 int
+			n2, s.err = s.w.Write(s.current)
+			if s.err != nil {
+				return s.err
+			}
+			s.nWritten += int64(n2)
+			s.current = s.current[:0]
+			s.filling = s.filling[:0]
+			return nil
+		}
```
vmihailenco: I'd rather not. This just hides the problem if I read this correctly.

klauspost: What?

klauspost: This is not hiding any problems, but handles small < 1 block streaming encodes.

vmihailenco: Yes, I understand. The benchmark will be faster until some threshold and then it will degrade to previous levels. So from my POV it does not change much and just gives false hopes :)

klauspost: When you are doing multiple blocks you will get the benefit of the concurrency. There will off course be a small dropoff as you get just above 1 block, but overall speed will be faster.

klauspost: But when we have enough for 1 block we don't know how much more will come, so we start compressing the first block.

vmihailenco: >but overall speed will be faster

Just want to repeat here as well that in case of HTTP server I don't want code to be faster by using more threads and CPU - those threads and CPU can be used to serve other requests/users. Otherwise I just make 1 request faster by delaying all the other requests.

klauspost: I respectfully disagree, but will also repeat that this will likely be the case when full parallel compression.

First of all Go has very advanced preemption and handles this nicely. 

Secondly you are just postponing the problem. If you have a 2 CPU system, the problem will occur as soon as you have 3 users and so on.

If you just make compression slower by using 1 of 2 cores, it will take (in theory) 2x as long, meaning you are strained for resources 2x as long instead of just getting it done faster AND give a better user experience when you are not limited for resources.

Assuming things scale linearly, you only can do xMB/s on your system and there is no reason to slow down every transaction, since the system is likely to behave the same when you reach that total limit anyway, ie. splitting the resources between the users.

vmihailenco: I mean that parallel version has some significant overhead. So instead of 100% CPU it uses 130% CPU and the 2nd user now has only 70% of the 2nd core.

What is even worse is that parallel version is slower on small payloads - not faster. I even tried to compress ~300mb binary and got this result:

```
BenchmarkCZstd-12      	       1	1758534769 ns/op	 160.09 MB/s	         0.239 compression	1288323136 B/op	      38 allocs/op
BenchmarkGoZstd-12     	       1	3465210251 ns/op	  81.24 MB/s	         0.244 compression	1339945080 B/op	     680 allocs/op
BenchmarkGoZstd2-12    	       1	2187101331 ns/op	 128.72 MB/s	         0.244 compression	757003880 B/op	     186 allocs/op
BenchmarkS2-12         	       2	 746165990 ns/op	 377.29 MB/s	         0.363 compression	1275335332 B/op	    1597 allocs/op
```

EncodeAll version is still significantly faster...

klauspost: You must be doing something strange. Here is a bench that doesn't allocate (as much): https://gist.github.com/klauspost/446e9ab7aeae0b75d7974339b65df815

```
λ go test -bench=. -benchtime=10s
BenchmarkCZstd-32          13430            889105 ns/op         253.44 MB/s          25 B/op          0 allocs/op
BenchmarkGoZstd-32          7183           1545090 ns/op         145.84 MB/s      251130 B/op          1 allocs/op
BenchmarkGoZstd2-32         5437           2187727 ns/op         103.00 MB/s      112531 B/op          0 allocs/op
BenchmarkS2-32             22752            528732 ns/op         426.19 MB/s         975 B/op          8 allocs/op
```

PR #262, opened 2020-05-26
Remove superfluous encoder.Reset()

dferstay:
When using the Apache Pulsar go client
(https://github.com/apache/pulsar-client-go) configured to use zstd
compression during message production in high-throughput service, we
see high-memory allocation due to Encoder.Reset().

Encoders on the `e.encoders` channel are already reset;
there is no need to reset an encoder after pulling it from the channel.

Signed-off-by: Daniel Ferstay <dferstay@splunk.com>

### DIFF
--- a/zstd/encoder.go
+++ b/zstd/encoder.go
@@ -82,7 +82,9 @@ func (e *Encoder) initialize() {
 	}
 	e.encoders = make(chan encoder, e.o.concurrent)
 	for i := 0; i < e.o.concurrent; i++ {
-		e.encoders <- e.o.encoder()
+		enc := e.o.encoder()
+		enc.Reset()
+		e.encoders <- enc
 	}
 }
 
@@ -448,7 +450,6 @@ func (e *Encoder) EncodeAll(src, dst []byte) []byte {
 		enc.Reset()
 		e.encoders <- enc
 	}()
-	enc.Reset()
 	blk := enc.Block()
 	// Use single segments when above minimum window and below 1MB.
 	single := len(src) < 1<<20 && len(src) > MinWindowSize


klauspost:
Very nice! Thanks for the contribution. 

I plan to make a release soon. Probably when #260 is done which is making slow but steady progress.

klauspost:
@dferstay Out of curiosity - where did you observe the allocations (code line or type of alloc)?

It shouldn't really do any allocations once it has 'warmed up', so I am a bit curious what parts are doing the allocations.

klauspost:
I will check the failures tomorrow, heading to bed now. It should work as you propose the change AFAICT. There might be a problem where it relies on the double Reset when re-using.

klauspost:
It appears to pick of phantom matches across Resets when only one is done.

dferstay:
> @dferstay Out of curiosity - where did you observe the allocations (code line or type of alloc)?
> 
> It shouldn't really do any allocations once it has 'warmed up', so I am a bit curious what parts are doing the allocations.

We see allocations coming from the `zstd/enc_fast.go`, the `Reset()` method.  Looking at the code in Reset(), there are attempts to reuse existing objects and the arrays that back slices, so I'm a little confused where the space is being burned.  See the attached pprof -inuse_space graph.

pprof_heap_alloc

dferstay:
We are currently using the default encoder options, which means we are using a doubleFastEncoder with:
- blockSize:  1 << 16
- windowSize: 8 << 20 
With the above windowSize, the encoder's `maxMatchOff` is ~33MB; when the histogram for the encoder is allocated it will have a capacity of 2*maxMatchOff, or ~66MB.  Is ~288MB normal memory use for a doubleFastEncoder?

klauspost:
Correct, you did the math yourself. To support the default number of concurrent encodes it needs a history for each. I am guessing you are testing on a CPU with `GOMAXPROCS` set to 8 and 33x8 = 264.

You can tweak the number of encoders available with `WithEncoderConcurrency`  (will enable more or less concurrent encodes) and the window size with `WithWindowSize` (smaller gives less compression).

I will look into the complexity of reducing the upfront allocations where we know the payload size when using EncodeAll. When it is less than the window size we don't need to track the history at all.

klauspost:
I have updated #263 to only allocate history when it is needed, so single blocks < 128K will not have a history allocated (but will keep it if it already has been).

I don't want to risk having to re-allocate history, so we either allocate what we want or nothing at all if we don't need it.

This will give double Resets in some cases, but that itself is pretty harmless.

To control maximum memory usage you can adjust the parameters described above.

dferstay:
> I am guessing you are testing on a CPU with GOMAXPROCS set to 8 and 33x8 = 264

Correct.

> I have updated #263 to only allocate history when it is needed, so single blocks < 128K will not have a history allocated (but will keep it if it already has been).

Excellent!

> To control maximum memory usage you can adjust the parameters described above.

Thank you very much for the explanation; much appreciated.

PR #266, opened 2020-06-07
zip: update CreateHeaderRaw to handle zip64 fields.

saracen:
This adds the zip64 field handling identical to [fileWriter.Close](https://github.com/klauspost/compress/blob/26bacdfece2923f6b0b4c7abb1a9527628c73c89/zip/writer.go#L612).

Without this, it is far easier to create an invalid ZIP file. This provides convenience for when you only want to set the compressed data length and CRC32 fields after using `zip.FileInfoHeader`, rather than handle multiple fields manually.

### DIFF
--- a/zip/writer.go
+++ b/zip/writer.go
@@ -218,7 +218,7 @@ func (w *Writer) Close() error {
 // allowed. To create a directory instead of a file, add a trailing
 // slash to the name.
 // The file's contents must be written to the io.Writer before the next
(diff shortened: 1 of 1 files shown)


klauspost:
Thanks for the contribution.

saracen:
@klauspost Thank you for implementing the `CreateHeaderRaw` proposal 👍

PR #280, opened 2020-08-31
zstd: Decoder: implement io.Closer

peterbourgon:
Permit zstd.Decoder to implement io.Closer without an adapter.


klauspost:
I will make this change and release a `v1.11`.  The breakage should be limited ;)

klauspost:
@peterbourgon  I forgot, there is the [(*Decoder).IOReadCloser()](https://pkg.go.dev/github.com/klauspost/compress@v1.10.11/zstd?tab=doc#Decoder.IOReadCloser) so there is no reason to break the API.

PR #291, opened 2020-10-27
s2: Fix "better" out of bounds read.

### LINKED ISSUE #290: s2.EncodeBetter reads out of bounds and panics
When I run

```go
package main

import (
	"io/ioutil"
	"log"
	"os"

	"github.com/klauspost/compress/s2"
)

func main() {
	p, err := ioutil.ReadFile(os.Args[1])
	if err != nil {
		log.Fatal(err)
	}
	p = p[:len(p):len(p)]

	s2.EncodeBetter(nil, p)
}
```

on the [attached file](https://github.com/klauspost/compress/files/5447579/s2fail.txt), I get

```
panic: runtime error: slice bounds out of range [:8] with capacity 6

goroutine 1 [running]:
github.com/klauspost/compress/s2.load64(...)
	/home/me/go/pkg/mod/github.com/klauspost/compress@v1.11.1/s2/encode_all.go:22
github.com/klauspost/compress/s2.encodeBlockBetter(0xc00011c003, 0x10011, 0x10011, 0xc00010a000, 0x1000d, 0x1000d, 0x0)
	/home/me/go/pkg/mod/github.com/klauspost/compress@v1.11.1/s2/encode_better.go:186 +0xc68
github.com/klauspost/compress/s2.EncodeBetter(0xc00011c000, 0x10014, 0x10014, 0xc00010a000, 0x1000d, 0x1000d, 0x0, 0xc000014118, 0x0)
	/home/me/go/pkg/mod/github.com/klauspost/compress@v1.11.1/s2/encode.go:93 +0x20d
main.main()
	/home/me/tmp/tests2.go:18 +0xd8
exit status 2
```

It looks like some bounds check is missing. The funny thing is that the program doesn't crash if I omit the `p = p[:len(p):len(p)]`, but the program were this crash first occurred didn't pull that trick.

klauspost:
Fixes #290


klauspost:
Travis seems to have broken its integration again: https://travis-ci.org/github/klauspost/compress/builds/739402780

greatroar:
Fix confirmed.

s2/encode_test.go
```diff
@@ -131,7 +145,7 @@ func TestEncoderRegression(t *testing.T) {
 				t.Error(err)
 				return
 			}
-			test(t, b)
+			test(t, b[:len(b):len(b)])
```
greatroar, line 148: Might it be a good idea to restrict the capacity of src in EncodeBetter, too? (And maybe other places?)

klauspost: @greatroar That is exactly what this is doing. The `test` function tests both Encode and EncodeBetter using the data.

klauspost:
> Fix confirmed. 

Great. I will wait until CI has finished and I also have started fuzz testing. If both are fine I will merge and release.

s2/encode_test.go
greatroar: I mean in the actual code, not in the test.

Issue #253, opened 2020-04-27
zstd: provide a "contains magic number" func


mvdan:
That is, to detect if a chunk of bytes has the zstd magic number as a prefix.

There are Go modules like https://github.com/h2non/filetype, which support lots of file types. But at the end of the day, I just need to check four bytes, and it seems unnecessary to add an extra module dependency when I already depend on a zstd package :)

For example, here's a sample implementation I wrote, which could be used like `if zstd.HasMagic(chunk) { ...`:

```
// HasMagic reports whether src begins with zstd v1's magic number,
// 0xFD2FB528 in little-endian format.
func HasMagic(src []byte) bool {
        return len(src) >= 4 &&
                src[0] == 0x28 && src[1] == 0xB5 &&
                src[2] == 0x2f && src[3] == 0xFD
}
```

I only raised this as an issue instead of a PR because the patch is tiny and I feel like there's a lot of room for bikeshedding, or discussion around whether or not we want this.

mvdan:
(But of course, happy to submit a PR if this seems like a good idea)

klauspost:
It could be part of #237 which would also check more frame header fields.

mvdan:
Is that from the point of view of exposing less API, or so that the function does extra sanity checks?

I can sympathize with exposing less API. I don't care about extra sanity checks though, as I directly follow up "is the magic number there" with "decompress the bytes". I guess I could ignore the rest of the data as long as it's equally cheap to retrieve.

klauspost:
Less API 💯

mvdan:
Sounds good to me :) I'll continue using a local helper while I wait for this issue to be fixed as part of #237 - there's no rush.

chrislusf:
Any update on this?

It would be great if the magic number can be prefixed without API changes.

PR #295, opened 2020-11-15
zstd: Tweak DecodeAll standard allocs

klauspost:
Default to compressed size * 2 if no frame size and no destination buffer given.

### DIFF
--- a/zstd/decoder.go
+++ b/zstd/decoder.go
@@ -323,18 +323,22 @@ func (d *Decoder) DecodeAll(input, dst []byte) ([]byte, error) {
 		}
 		if frame.FrameContentSize > 0 && frame.FrameContentSize < 1<<30 {
 			// Never preallocate moe than 1 GB up front.
-			if uint64(cap(dst)) < frame.FrameContentSize {
+			if cap(dst)-len(dst) < int(frame.FrameContentSize) {
 				dst2 := make([]byte, len(dst), len(dst)+int(frame.FrameContentSize))
 				copy(dst2, dst)
 				dst = dst2
 			}
 		}
 		if cap(dst) == 0 {
-			// Allocate window size * 2 by default if nothing is provided and we didn't get frame content size.
-			size := frame.WindowSize * 2
+			// Allocate len(input) * 2 by default if nothing is provided
+			// and we didn't get frame content size.
+			size := len(input) * 2
 			// Cap to 1 MB.
 			if size > 1<<20 {
 				size = 1 << 20
+			}
+			if uint64(size) > d.o.maxDecodedSize {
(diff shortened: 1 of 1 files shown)


zstd/decoder.go
```diff
@@ -323,19 +323,23 @@ func (d *Decoder) DecodeAll(input, dst []byte) ([]byte, error) {
...
 				copy(dst2, dst)
 				dst = dst2
 			}
 		}
 		if cap(dst) == 0 {
-			// Allocate window size * 2 by default if nothing is provided and we didn't get frame content size.
-			size := frame.WindowSize * 2
+			// Allocate len(input) * 2 by default if nothing is provided
+			// and we didn't get frame content size.
+			size := len(input) * 2
```
larry-cdn77, line 335: Is this likely to lead to the sequence decoder reallocating to 2MB maxBlockSize?

klauspost: @larry-cdn77 The 1MB cap check is still, see next lines.

klauspost: @larry-cdn77 And you can always just allocate with any custom size and send that as the destination.

larry-cdn77: Certainly. Thank you for responding.

I wonder if I am right to observe that in certain situations one has to hit a sweetspot in the size of custom slice to pass in. I have a high throughput application with message sizes of less than 1KB. The custom slice passed in has to be relatively small to avoid garbage collection spinning up to a lot of CPU time (I've seen a 50% throughput penalty). And it has to big enough to stay clear of sequence decoder [replacing](https://github.com/klauspost/compress/blob/ff4a4acf3be7ac7b0b1c3381000e0216eb9e956f/zstd/seqdec.go#L184) it with a new 2MB one, again degrading performance.

Currently the size of my slice is a multiple of compressed size, and experimentally a multiplier of 32 appears to be the sweetspot (I didn't fully understand if this is how much compression my encoder has achieved, or there are other factors behind that number).

Related to this, would it make sense to adjust the library's maxBlockSize given that my data is zstd library encoded and I believe they use 128KB maximum block size?

klauspost: @larry-cdn77 If you are using this package to compress it doesn't matter, since it will have FrameContentSize which allows to make an exact allocation and this will never be relevant.

>  a multiplier of 32

So you have a 32:1 compression ratio? That is way above what most will see, so I can't use that metric. With an expected factor of 2 we should be able to decode most content with 1 or 2 allocations.

`maxBlockSize` is defined by the format, so it cannot be changed. This is the maximum size of an uncompressed block.

We could take a look at https://github.com/klauspost/compress/issues/253 which could be made to decode the frame and first block headers and return the size if there is only one block.

Issue #296, opened 2020-11-21
zstd: Decoder.Reset should allow nil input


klauspost:
The decoder should allow nil input to clear references to the previous stream.

PR #305, opened 2020-12-25
zstd: free Decoder resources when Reset is called with a nil io.Reader

mostynb:
Attempt at fixing #296.

### DIFF
--- a/zstd/decoder.go
+++ b/zstd/decoder.go
@@ -156,8 +156,14 @@ func (d *Decoder) Reset(r io.Reader) error {
 	if d.current.err == ErrDecoderClosed {
 		return d.current.err
 	}
+
+	d.drainOutput()
+
 	if r == nil {
-		return errors.New("nil Reader sent as input")
+		d.current.b = []byte{}
+		d.current.err = io.EOF
+		d.current.flushed = true
+		return nil
 	}
 
 	if d.stream == nil {
@@ -165,8 +171,6 @@ func (d *Decoder) Reset(r io.Reader) error {
 		d.streamWg.Add(1)
 		go d.startStreamDecoder(d.stream)
 	}
-
-	d.drainOutput()
 
 	// If bytes buffer and < 1MB, do sync decoding anyway.
 	if bb, ok := r.(*bytes.Buffer); ok && bb.Len() < 1<<20 {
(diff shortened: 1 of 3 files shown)


mostynb:
(I will dig into the test failures in the next few days.)

zstd/decoder.go
```diff
@@ -156,8 +156,14 @@ func (d *Decoder) Reset(r io.Reader) error {
 	if d.current.err == ErrDecoderClosed {
 		return d.current.err
 	}
+
+	d.drainOutput()
+
 	if r == nil {
-		return errors.New("nil Reader sent as input")
+		d.current.b = []byte{}
```
klauspost, line 163: This should already be handled by the `drainOutput`, so this should be an un-needed alloc.
```suggestion
```

zstd/decoder.go
```diff
@@ -156,8 +156,14 @@ func (d *Decoder) Reset(r io.Reader) error {
...
+		d.current.err = io.EOF
```
klauspost, line 164: This would make a read seem like the payload was a valid 0 byte block. 

We want to make it clear that a read was performed with a nil input.

```Go
	// ErrNilInput is returned when a nil Reader was given
	// and a non-Reset operation was attempted. 
	ErrNilInput = errors.New("nil input provided as reader")
```

```suggestion
		d.current.err = ErrNilInput
```

klauspost (commented):
Thanks for the contribution.

Let's update the docs as well.
```Go
// Reset will reset the decoder the supplied stream after the current has finished processing.
// Note that this functionality cannot be used after Close has been called.
// Reset can be called with a nil reader to release references to the previous reader.
// When sending a nil reader no other operations than Reset or DecodeAll or Close should be used.
func (d *Decoder) Reset(r io.Reader) error {
```

A test that also tests the basic functionality, ie create a Reader, Reset with nil and check that reads returns the expected error and finally that it can be used with a proper reader would be great.

zstd/decoder.go
mostynb: I guess we should also return this sentinel error if the Decoder was created with a nil reader.

zstd/decoder.go
mostynb: Removed in the fixup.

zstd/decoder.go
mostynb: Added this change, but with the name ErrDecoderNilInput in the fixup. Maybe that's too specific though?

klauspost: It's fine. Thanks!

klauspost:
Thanks. Looks good. I will give it a final look through and merge when I have the time.

PR #308, opened 2021-01-07
s2: Fix crash on small better buffers

klauspost:


klauspost:
Requires https://github.com/klauspost/compress/pull/308 to pass.

PR #309, opened 2021-01-07
Use interface to get bytes

klauspost:
Use interface to detect bytes.Byffer.

Move generator.


klauspost:
Requires https://github.com/klauspost/compress/pull/308 to pass.

PR #339, opened 2021-03-19
Fix a few unnecessary type conversions

Jacalz:
This PR gets rid of a bunch of unneeded type conversions. I was checking out the code and
saw a few of these myself, but most of them were fixed using [unconvert](https://github.com/mdempsky/unconvert).


klauspost:
@Jacalz Thanks!

PR #342, opened 2021-03-20
Fix some minor issues reported by staticcheck

Jacalz:
This just fixes a few minor issues reported by staticcheck. Nothing major.

I did not know how to solve these, but I expect that they may be of interest for performance reasons:
```
/home/jacob/Git/compress/s2/encode.go
  (402, 19)  SA6002  argument should be pointer-like to avoid allocations
  (655, 18)  SA6002  argument should be pointer-like to avoid allocations
  (734, 17)  SA6002  argument should be pointer-like to avoid allocations
  (819, 17)  SA6002  argument should be pointer-like to avoid allocations
```

### DIFF
--- a/flate/deflate_test.go
+++ b/flate/deflate_test.go
@@ -100,9 +100,9 @@ func TestBulkHash4(t *testing.T) {
 						t.Errorf("Len:%d Index:%d, expected 0x%08x but not modified", len(y), i, expect)
 					} else if got != expect {
 						t.Errorf("Len:%d Index:%d, got 0x%08x expected:0x%08x", len(y), i, got, expect)
-					} else {
-						//t.Logf("Len:%d Index:%d OK (0x%08x)", len(y), i, got)
-					}
+					} /*else {
+						t.Logf("Len:%d Index:%d OK (0x%08x)", len(y), i, got)
+					} */
 				}
 			}
 		}
@@ -561,7 +561,7 @@ func testResetOutput(t *testing.T, name string, newWriter func(w io.Writer) (*Wr
 		if len(out1) != len(out2) {
 			t.Errorf("got %d, expected %d bytes", len(out2), len(out1))
 		}
-		if bytes.Compare(out1, out2) != 0 {
+		if !bytes.Equal(out1, out2) {
 			mm := 0
 			for i, b := range out1[:len(out2)] {
 				if b != out2[i] {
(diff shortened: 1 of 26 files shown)


klauspost:
It wants to have the byte buffers as `*[]byte` to avoid 24bytes of alloc when not reused.

I don't want to uglify the code to cater for that check.

flate/deflate_test.go
```diff
@@ -100,9 +100,9 @@ func TestBulkHash4(t *testing.T) {
 						t.Errorf("Len:%d Index:%d, expected 0x%08x but not modified", len(y), i, expect)
 					} else if got != expect {
 						t.Errorf("Len:%d Index:%d, got 0x%08x expected:0x%08x", len(y), i, got, expect)
-					} else {
-						//t.Logf("Len:%d Index:%d OK (0x%08x)", len(y), i, got)
-					}
+					} /*else {
+						t.Logf("Len:%d Index:%d OK (0x%08x)", len(y), i, got)
+					} */
```
klauspost, line 105: Revert please.

gzip/gunzip_test.go
```diff
@@ -577,9 +575,6 @@ func TestTruncatedStreams(t *testing.T) {
 			continue
 		}
 		_, err = io.Copy(ioutil.Discard, r)
-		if ferr, ok := err.(*flate.ReadError); ok {
-			err = ferr.Err
-		}
```
klauspost, line 582: What is wrong with this unwrap?

klauspost (commented):
LGTM, except some minor stuff.

gzip/gunzip_test.go
Jacalz: It is deprecated and the deprecation notice states that it is not returned any more. https://github.com/klauspost/compress/blob/d693bc87c5fbde6c49b19afaecdddf3e369389c7/flate/inflate.go#L55-L65 A quick search of GitHub confirms that it is correct.

flate/deflate_test.go
Jacalz: Fixed in the latest commit.

Jacalz:
Sorry for this taking quite some time to get to. I did not have access to my computer this last week. Should be ready for a re-review now though :)

klauspost:
Thanks a bunch!

PR #343, opened 2021-03-20
Update vendored xxhash to 1a548c8655f85f3ffcf3578aa1dc8f922ecb6a98

Jacalz:
There are two changes of note here. Firstly there is https://github.com/cespare/xxhash/commit/fd5c34b71e8ec6a322b55ae01fde54e8edea759a that removes a useless assignment. More importantly, there is also https://github.com/cespare/xxhash/commit/a7909af24e080cb18e105af79be6915531147a86 which allows some parts of the code to be inlined for between 10 to 30 percent speed improvements.

### DIFF
--- a/zstd/internal/xxhash/xxhash_safe.go
+++ b/zstd/internal/xxhash/xxhash_safe.go
@@ -1,3 +1,7 @@
+// +build appengine
+
+// This file contains the safe implementations of otherwise unsafe-using code.
+
 package xxhash
 
 // Sum64String computes the 64-bit xxHash digest of s.
--- a/zstd/internal/xxhash/xxhash_unsafe.go
+++ b/zstd/internal/xxhash/xxhash_unsafe.go
@@ -1,0 +1,57 @@
+// +build !appengine
+
+// This file encapsulates usage of unsafe.
+// xxhash_safe.go contains the safe implementations.
+
+package xxhash
+
+import (
+	"unsafe"
+)
+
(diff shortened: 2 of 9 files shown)


zstd/internal/xxhash/xxhash_unsafe.go
```diff
@@ -0,0 +1,57 @@
+// +build !appengine
```
klauspost, line 1: Remove the unsafe package stuff. It is not needed and I don't want to import unsafe.

zstd/internal/xxhash/xxhash_safe.go
```diff
@@ -1,3 +1,7 @@
+// +build appengine
+
+// This file contains the safe implementations of otherwise unsafe-using code.
+
```
klauspost, line 4:
```suggestion
```

zstd/internal/xxhash/xxhash_unsafe_test.go
```diff
@@ -0,0 +1,61 @@
+// +build !appengine
```
klauspost, line 1: Remove file.

klauspost (commented):
Please remind me why we update this? There is no functional change AFAICT.

Jacalz:
> Please remind me why we update this? There is no functional change AFAICT.

You're right- Without using the unsafe stuff, this is pretty much useless.

PR #363, opened 2021-04-22
zstd: Add helpers to compress/decompress zstd inside zip files

klauspost:
Adds ZipCompressor and ZipDecompressor as well as method constants.

Winzip confirmed... PKWARE, no idea... download form doesn't work.

7zip detects it, but cannot decompress.

### DIFF
--- a/zstd/zip.go
+++ b/zstd/zip.go
@@ -1,0 +1,120 @@
+// Copyright 2019+ Klaus Post. All rights reserved.
+// License information can be found in the LICENSE file.
+
+package zstd
+
+import (
+	"errors"
+	"io"
+	"sync"
+)
+
+// ZipMethodWinZip is the method for Zstandard compressed data inside Zip files for WinZip.
+// See https://www.winzip.com/win/en/comp_info.html
+const ZipMethodWinZip = 93
+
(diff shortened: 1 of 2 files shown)


romiras:
@klauspost
FYI, a compression method 20 declared as deprecated in favor of 93 since Zip format spec. version 6.3.8 (currently 6.3.**9**)
You can see in section 4.4.5:
```
       20 - deprecated (use method 93 for zstd)
       93 - Zstandard (zstd) Compression 
```

Therefore these two constants can be merged into one:
```go
// ZipMethodWinZip is the method for Zstandard compressed data inside Zip files for WinZip.
// See https://www.winzip.com/win/en/comp_info.html
const ZipMethodWinZip = 93

// ZipMethodPKWare is the method number used by PKWARE to indicate Zstandard compression.
// See https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.7.TXT
const ZipMethodPKWare = 20
```


https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.9.TXT

klauspost:
@romiras Thanks for the ping. Adding a note in https://github.com/klauspost/compress/pull/369 - it might be useful in rare cases, so I kept the constant.

PR #373, opened 2021-05-11
zstd: Forward read errors

### LINKED ISSUE #372: zstd reader does not transmit upstream reader failure
If the underlying reader (the one given to zstd.NewReader) fails and returns an error, zstd just returns EOF instead of transmitting the error : 

```go
package main

import (
	"fmt"
	"log"	
	"io"
)

type errorReader struct{}

func (r *errorReader) Read(p []byte) (int, error) {
	return 0, fmt.Errorf("i'm a failure")
}

func main() {
	er := new(errorReader)
	zr, err := zstd.NewReader(er)	
	if err != nil {
		log.Fatal(err)
	}
	
	_, err = io.ReadAll(zr)
	if err != nil {
		log.Fatal(err)
	}
	
	log.Print("reached eof without any error !")
}
```

Expected result : "I'm a failure"
Actual result: "reached eof without any error !"

klauspost:
Forward all read errors.

Fixes #372

### DIFF
--- a/zstd/blockdec.go
+++ b/zstd/blockdec.go
@@ -123,12 +123,10 @@ func newBlockDec(lowMem bool) *blockDec {
 // Input must be a start of a block and will be at the end of the block when returned.
 func (b *blockDec) reset(br byteBuffer, windowSize uint64) error {
 	b.WindowSize = windowSize
-	tmp := br.readSmall(3)
-	if tmp == nil {
-		if debug {
-			println("Reading block header:", io.ErrUnexpectedEOF)
-		}
-		return io.ErrUnexpectedEOF
+	tmp, err := br.readSmall(3)
+	if err != nil {
+		println("Reading block header:", err)
+		return err
 	}
 	bh := uint32(tmp[0]) | (uint32(tmp[1]) << 8) | (uint32(tmp[2]) << 16)
 	b.Last = bh&1 != 0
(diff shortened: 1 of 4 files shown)


zstd/blockdec.go
```diff
@@ -123,12 +123,10 @@ func newBlockDec(lowMem bool) *blockDec {
...
 	b.WindowSize = windowSize
-	tmp := br.readSmall(3)
-	if tmp == nil {
-		if debug {
-			println("Reading block header:", io.ErrUnexpectedEOF)
-		}
-		return io.ErrUnexpectedEOF
+	tmp, err := br.readSmall(3)
+	if err != nil {
+		println("Reading block header:", err)
```
twmb, line 128: Are these printlns meant to be gated with `debug`?

klauspost: @twmb There is another check in the [println](https://github.com/klauspost/compress/blob/a5aad11bec853fa7441dfa47cf5b3e05b7ede0e4/zstd/zstd.go#L91) function, so it will not print unless compiled with debug enabled. 

After encoding and decoding was separated it should, though.

twmb: Ah thank you very much! I didn't see that this was an override of the Go builtin, but seeing two arguments should've been the hint.

PR #388, opened 2021-06-04
zip: only preallocate File slice if reasonably sized

klauspost:
Since the number of files in the EOCD record isn't validated, it isn't
safe to preallocate Reader.Files using that field. A malformed archive
can indicate it contains up to 1 << 128 - 1 files. We can still safely
preallocate the slice by checking if the specified number of files in
the archive is reasonable, given the size of the archive.

Thanks to the OSS-Fuzz project for discovering this issue and to
Emmanuel Odeke for reporting it.

See https://github.com/golang/go/issues/46242


klauspost:
Not needed check was added in #13

PR #391, opened 2021-06-08
Add ability to decompress streams created with very large windows (--long=30, --long=31)

pushshift:
The zstd compression program allows for compression with window sizes up to 2 << 31. This pull request adjusts the constant for MaxWindowSize to accommodate larger window sizes.

I do not believe this change would break any existing code that uses this module.


pushshift:
Made an editing error for some reason on one of the files. Closing pull request for the time being. My bad.

PR #403, opened 2021-07-03
Fixed decoder Close() return

CalebQ42:
I noticed that zstd.Decoder's Close function was missing the error return necessary for io.Closer.


klauspost:
By design. Use https://pkg.go.dev/github.com/klauspost/compress/zstd#Decoder.IOReadCloser

PR #416, opened 2021-08-16
tests: listen on “localhost” instead of IPv4

stapelberg:
This makes the test work in IPv6-only environments, too.


klauspost:
@stapelberg Thanks for the contribution. LGTM. Merging when tests are done.

klauspost:
(renamed for changelog neatness)

PR #420, opened 2021-08-19
zstd: Minor performance tweaks

klauspost:
These seem better in Go 1.17


odeke-em:
Nice performance boosts @klauspost https://dashboard.github.orijtech.com/benchmark/d2e6bf89f9504259a09c6ab891049cc1

klauspost:
@odeke-em Yeah, unfortunately it is likely a regression on Go1.16-

The Go 1.17 compiler has quite different "best-case" input.

PR #426, opened 2021-08-30
zstd: pooledZipWriter should return Writers to the same pool

mostynb:
This code was previously Get'ing objects from one pool and later Put'ing them to a different pool, so they were never reused.

### DIFF
--- a/zstd/zip.go
+++ b/zstd/zip.go
@@ -64,8 +64,9 @@ func (r *pooledZipReader) Close() error {
 }
 
 type pooledZipWriter struct {
-	mu  sync.Mutex // guards Close and Read
-	enc *Encoder
+	mu   sync.Mutex // guards Close and Read
+	enc  *Encoder
+	pool *sync.Pool
 }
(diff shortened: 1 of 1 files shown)


mostynb:
Also, `pooledZipReader` seems to be unused because it leaks goroutines- would you be interested in merging something like https://github.com/mostynb/zstdpool-syncpool/blob/main/decoderpool.go ?

klauspost:
@mostynb Nice! Thanks!

I don't want to rely on finalizers. I want to fix it properly, ie have a streaming option that only uses the main goroutine.

mostynb:
> I don't want to rely on finalizers. I want to fix it properly, ie have a streaming option that only uses the main goroutine.

I agree that would be preferable. Is it worth commenting out pooledZipReader in the meantime, since it's unused?

klauspost:
@mostynb It is fine to leave. Thanks for the contribution! I will include it in the release I will do shortly.

PR #427, opened 2021-08-30
Make zstd compatible with io.Closer() interface

darkdragon-001:
Make `zstd.Decoder` compatible with [`io.Closer()`](https://pkg.go.dev/io#Closer) interface.


klauspost:
No breaking API changes: Use [Decoder.IOReadCloser](https://pkg.go.dev/github.com/klauspost/compress/zstd#Decoder.IOReadCloser)

PR #432, opened 2021-09-01
zip: Add fs.FS support for Go 1.16+

klauspost:
Change writer api to match stdlib as well.

`func (w *Writer) Copy(name string, src *File) error` is changed to `func (w *Writer) Copy(src *File) error` to match the stdlib.


odeke-em:
Well done @klauspost, improvements for the benchmarks you care about https://dashboard.github.orijtech.com/benchmark/c099d0acb095438bb5590e89b746a405 :-) 

If I may kindly ask, for notifications, would you like a comment for every commit telling you when it completed? Or would you prefer the Github Actions style notification that you have to proactively look at for completion on the PR's page? /cc @cuonglm @kirbyquerby

klauspost:
@odeke-em Yeah. This seems like a good example of non-changes yielding rather big fluctuations. None of the benchmarks are affected by the code change in this PR. Kind of frustrating that the Go compiler has so big variance in compiles.

I definitely would like a notification. If possible with a table. Some systems like this will delete previous results when results from a new push arrives. That would be great so history doesn't get too cluttered with outdated benchmarks.

/cc @cuonglm @kirbyquerby

saracen:
```
% go version 
go version go1.13.8 darwin/amd64
```
```
% go mod tidy
...
	github.com/klauspost/compress/zip imports
	io/fs: malformed module path "io/fs": missing dot in first path element
```

It looks like `go mod tidy` doesn't take into consideration build tags. This is a problem for us, as we use `go mod tidy` on the CI to ensure consistency.

@klauspost I don't expect there's anything that can be done about this, but figured I'd leave this here for anybody else that stumbles across this problem.

klauspost:
@saracen Minimum supported Go version is currently Go 1.15 - could you check on that?

saracen:
@klauspost That does appear to work. I figured the same thing would happen as `io/fs` was only introduced in go1.16. Apologies for the noise!

PR #434, opened 2021-09-01
Add performance baseline

klauspost:
Disables stdlib tests.

### DIFF
--- a/gzip/gzip_test.go
+++ b/gzip/gzip_test.go
@@ -7,7 +7,6 @@ package gzip
 import (
 	"bufio"
 	"bytes"
-	oldgz "compress/gzip"
 	"io"
 	"io/ioutil"
 	"math/rand"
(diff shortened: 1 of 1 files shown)


odeke-em:
@klauspost sorry for the confusion and thank you for using bencher! So after you merge this master, we'll finally have some baseline comparison values and then subsequent PRs will use that baseline :-) cc @cuonglm

klauspost:
@odeke-em Yeah, there are literally hundreds of benchmarks in this package, so your system will get quite a workout.

Maybe I should make the non-important benchmarks be ignored by `-test.short`, so I can filter them a bit.

odeke-em:
Thanks for the notice @klauspost! Good point about using `if testing.Short()`  As for the number of tests, it might take like 30+ minutes but the tests are running for example https://dashboard.github.orijtech.com/benchmark/f47cd13a3f1b4c5da331880af8f63adc

odeke-em:
@klauspost, now that you have the baseline data available, things should run smoothly. I also created my own fork of this repo and firstly started with a change in March 2020 https://github.com/orijtech/compress/commit/badde5230a5d116220c8066bc77f97905bb15957 and from that used the current code on your master branch and this was the interesting result https://dashboard.github.orijtech.com/benchmark/ddfc52a0d5a743cf8b0f3785fb5e1ef0

Lots of improvements in pretty much all the benchmarks. However, some like RandomEncode, GzipL6, Encoder*, Deflate/*, Decompress* have regressed quite a lot since that commit in 2020
Screen Shot 2021-09-01 at 11 44 11 AM


cc also @kirbyquerby

PR #435, opened 2021-09-02
huff0: Faster 1X Decompression

klauspost:
Go 1.17 compiler is weird.

### DIFF
--- a/huff0/decompress.go
+++ b/huff0/decompress.go
@@ -914,7 +914,7 @@ func (d *Decoder) decompress4X8bit(dst, src []byte) ([]byte, error) {
 	out := dst
 	dstEvery := (dstSize + 3) / 4
 
-	shift := (64 - d.actualTableLog) & 63
+	shift := (8 - d.actualTableLog) & 7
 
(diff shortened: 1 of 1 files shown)


odeke-em:
You have some regressions, am not sure if you intended them or not https://dashboard.github.orijtech.com/benchmark/da0d1c1df2a54d3d8ffa6362d8cbc53e

We are working on making the bot an action so that it'll display the status for every commit and you can make required checks instead of manually inspecting always :-)

klauspost:
@odeke-em All of these benchmarks are 100% unrelated. Most are compression benchmarks and this only changes decompression.

So I guess this gives an idea of the run-to-run variance. I am seeing the same locally with rather large compile-to-compile variance, especially after upgrading to Go 1.17. I know there isn't much you can do about that.

All of the CompressAllocations/EncodeAllDict65536_0/Encoder_EncodeAllHTML/WriterRandom/TwainEncode1e5/RandomEncodeBlock1MB/RandomEncodeBetterBlock16MB are in unrelated code, and it is of course sad that variance is >10% in some cases.

Issue #459, opened 2021-12-01
Expose skippable frames via the API.


rouzier:
Is it possible to add an API to use skippable frames.

https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#skippable-frames

klauspost:
@rouzier It is possible. You can of course prepend or append it to any encode.

Adding one is fairly simple. This is used for adding the final padding, and a standalone version could look like this:

```
// CreateSkippableFrame will add a skippable frame with the specified payload
// and append it to dst.
func CreateSkippableFrame(dst []byte, payload []byte) ([]byte, error) {
	if len(payload) == 0 {
		return dst, nil
	}
	if len(payload) > math.MaxUint32 {
		return dst, fmt.Errorf("requested skippable frame size (%d) > max uint32", len(payload))
	}
	dst = append(dst, 0x50, 0x2a, 0x4d, 0x18)
	f := uint32(len(payload))
	dst = append(dst, uint8(f), uint8(f>>8), uint8(f>>16), uint8(f>>24))
	dst = append(dst, payload...)
	return dst, nil
}
```
(untested, beware)

klauspost:
Adding functionality to include it should also have callbacks for getting the payloads when decoding.

PR #476, opened 2022-01-25
zstd: improve header decoder

dsnet:
Changes made:
* Add Header.HeaderSize to track the actual size of the header.
The header is variable length, so providing this information allows
the caller to do other low-level tasks, such as determining whether
the entirety of a zstandard frame is a single segment block.
* Add Header.Skippable{ID,Size} to make the Header type more
useful for extracting data from metadata frames (see #459).
* Re-arrange fields in Header to better match the order
these appear in RFC 8478.


klauspost (approved):
Thanks for the contribution. LGTM

PR #501, opened 2022-02-21
archive/zip: don't read data descriptor early

saracen:
This is a patch originally applied to stdlib's `archive/zip`: https://go-review.googlesource.com/c/go/+/357489/

Your own `compress/zip` fork synced with stdlib after the problem was introduced, but before this commit, which reverted the change.


klauspost:
@saracen Why is the test removed?

saracen:
> @saracen Why is the test removed?

@klauspost This test was added along with the modified `readDataDescriptor` that has now been reverted.

The test was removed, because like the modified `readDataDescriptor`, it was redundant. The data it unmarshaled into the `dataDescriptor` struct was not used. The CRC32 value looks like it was used, but served no purpose, as it wasn't then checked against (something the older implementation and the reverted commit do).

klauspost:
Great! Thanks!

klauspost:
I expect a release soon to include #503 before merging the zstd decoder changes.

PR #502, opened 2022-02-21
Update deprecated CreateHeaderRaw to correctly call CreateRaw

saracen:
[fastzip](https://github.com/saracen/fastzip) had reports of malformed compressed data (https://github.com/saracen/fastzip/issues/32). I've narrowed the fix down to this.

I'll update `fastzip` to call the non-deprecated function now, but fixing this here is still a good idea. [gitLab-runner ](https://gitlab.com/gitlab-org/gitlab-runner) uses fastzip and stumbled across this issue when the `klauspost/compress` dependency was indirectly updated via a minor minio upgrade. 😢


klauspost:
Thanks for the contrib.

PR #524, opened 2022-03-10
tests: [zstd] Factor out common benchmark routine

WojciechMula:
Factored out common code for benchmark decompression of files. This is used to benchmark against the Silesia corpus and Enwik9; I corrected the path from `enwik8.zst` to `enwik9.zst`.

I found this procedure useful when benchmarking against custom test files.


klauspost (approved):
👍🏼

PR #525, opened 2022-03-10
tests: [zstd] Unit tests for sequenceDecs.adjustOffset

WojciechMula:
I wrote these unit tests to validate an assembler version of the code. The procedure is non-trivial and I guess it might be handy to have tests for it. I know that the procedure is inlined in some places, but anyway.


klauspost (approved):
Thanks! LGTM. Merging when passed.

PR #533, opened 2022-03-15
cpuinfo: get x86 features in runtime

WojciechMula:
For now, we need to know whether an x86 CPU supports the BMI extensions.

I made it a separate module, as it may be used not only in `zstd`.

I tested the code manually, but TBH I don't know how to unit test this.


klauspost (changes requested):
Let move this to `internal/cpuinfo` so we don't have expose it as a feature.

WojciechMula:
> Let move this to `internal/cpuinfo` so we don't have expose it as a feature.

Moved.

klauspost (approved):

PR #545, opened 2022-03-31
zstd: asm version of decodeSync

WojciechMula:
A little hacking in the current generator allowed us to reuse almost all code. That's nice!

Part of #515.

For now `go test -run TestDecoder` pass, I'm working on fixing the remaining tests. Another thing is that PR does not incorporate yet the history support (waiting for #542)

Benchmark results from an Ice Lake machine. Some quite good speed ups are there!

```
benchmark                                                                 old ns/op     new ns/op     delta
BenchmarkDecoder_DecoderSmall/kppkn.gtb.zst-16                            4948849       3115659       -37.04%
BenchmarkDecoder_DecoderSmall/geo.protodata.zst-16                        884190        525662        -40.55%
BenchmarkDecoder_DecoderSmall/plrabn12.txt.zst-16                         16820567      13477758      -19.87%
BenchmarkDecoder_DecoderSmall/lcet10.txt.zst-16                           12468563      9806714       -21.35%
BenchmarkDecoder_DecoderSmall/asyoulik.txt.zst-16                         3829515       1815824       -52.58%
BenchmarkDecoder_DecoderSmall/alice29.txt.zst-16                          5235045       2684060       -48.73%
BenchmarkDecoder_DecoderSmall/html_x_4.zst-16                             1494667       1004488       -32.80%
BenchmarkDecoder_DecoderSmall/paper-100k.pdf.zst-16                       226176        180516        -20.19%
BenchmarkDecoder_DecoderSmall/fireworks.jpeg.zst-16                       126631        125630        -0.79%
BenchmarkDecoder_DecoderSmall/urls.10K.zst-16                             14001037      12040286      -14.00%
BenchmarkDecoder_DecoderSmall/html.zst-16                                 1013496       580645        -42.71%
BenchmarkDecoder_DecoderSmall/comp-data.bin.zst-16                        78160         62866         -19.57%
BenchmarkDecoder_DecodeAll/kppkn.gtb.zst-16                               592552        291325        -50.84%
BenchmarkDecoder_DecodeAll/geo.protodata.zst-16                           106829        62462         -41.53%
BenchmarkDecoder_DecodeAll/plrabn12.txt.zst-16                            1901163       902247        -52.54%
BenchmarkDecoder_DecodeAll/lcet10.txt.zst-16                              1417726       673019        -52.53%
BenchmarkDecoder_DecodeAll/asyoulik.txt.zst-16                            475107        223363        -52.99%
[...]

PR #548, opened 2022-04-04
zstd: Store previous offsets in registers

klauspost:
Not super visible in microbenches (presumably because of cache), but very visible in streaming decodes.

Before/after:

```
BenchmarkDecoderSilesia-32    	       4	 281168525 ns/op	 753.81 MB/s	   50218 B/op	      44 allocs/op
BenchmarkDecoderSilesia-32    	       6	 190584583 ns/op	1112.09 MB/s	   49446 B/op	      45 allocs/op

BenchmarkDecoderEnwik9-32    	       1	1439964200 ns/op	 694.46 MB/s	   71952 B/op	      51 allocs/op
BenchmarkDecoderEnwik9-32    	       1	1184307200 ns/op	 844.38 MB/s	   72144 B/op	      53 allocs/op
```

### DIFF
--- a/zstd/_generate/gen.go
+++ b/zstd/_generate/gen.go
@@ -146,6 +146,16 @@ func (o options) genDecodeSeqAsm(name string) {
 	mlP := Mem{Base: seqBase, Disp: 1 * 8} // Pointer to current ml
 	llP := Mem{Base: seqBase, Disp: 0 * 8} // Pointer to current ll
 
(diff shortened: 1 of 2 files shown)


WojciechMula:
> @WojciechMula Do you have objections to this? I can hold this off if it interferes with #545

Great! Please merge, there are no problems with rebasing.

Issue #561, opened 2022-04-21
asm panic: runtime error: index out of range


klauspost:
From fuzz tests:
```
panic: runtime error: index out of range [466] with length 466

goroutine 30 [running]:
github.com/klauspost/compress/zstd.(*sequenceDecs).decode(0xc000089e60, {0xc0003b9000, 0x1d2, 0x21d860})
	e:/gopath/src/github.com/klauspost/compress/zstd/seqdec_amd64.go:225 +0x99d
github.com/klauspost/compress/zstd.(*blockDec).decodeSequences(0xc00010c900, 0xc000089e58)
	e:/gopath/src/github.com/klauspost/compress/zstd/blockdec.go:679 +0x28d
github.com/klauspost/compress/zstd.(*Decoder).startStreamDecoder.func2()
	e:/gopath/src/github.com/klauspost/compress/zstd/decoder.go:721 +0x5f8
created by github.com/klauspost/compress/zstd.(*Decoder).startStreamDecoder
	e:/gopath/src/github.com/klauspost/compress/zstd/decoder.go:686 +0x27d
```
Sample: [002135096346d2e9c82e7b0b5ef85bacd61ddc17.gz](https://github.com/klauspost/compress/files/8532247/002135096346d2e9c82e7b0b5ef85bacd61ddc17.gz)

klauspost:
@WojciechMula

klauspost:
Maybe: `i := len(seqs) - ctx.iteration - 1`?

WojciechMula:
I'm looking at this, please assign me.

klauspost:
I have added the change to #559 for now.

PR #564, opened 2022-04-22
Add regression test for #561

WojciechMula:
@klauspost  This is what I had done before you fixed the issue. :) I'm wondering if there's an easy way to create arbitrary sequences to test `decode` more thoroughly.


klauspost (approved):

WojciechMula:
I just replaced generic `interface{}` with `testing.TB` interface and get rid of a typed case.

PR #567, opened 2022-04-24
zstd: Add benchmark for custom files

WojciechMula:
It's for users who want to benchmark against their own files.

An expected invocation is: `go test -run XYZ -bench BenchmarkDecoderWithCustomFiles`


klauspost:
👍🏼

PR #569, opened 2022-04-25
Fix incorrectly rejected blocks

klauspost:
Assembly would reject blocks if an exact buffer was provided and no extra literals.

```
        --- FAIL: TestEncoderRegression/best-c2-w16384k-1seg/054917b2373a9ca83257108e02e9c356e49d126f (0.02s)
            encoder_test.go:257: error: output (40533) bigger than max block size (131072)
                want: 829232
                got:  786432
```


zstd/seqdec_amd64.s
```diff
@@ -1913,6 +1913,7 @@ error_not_enough_space:
 	MOVQ CX, 208(AX)
 	MOVQ 16(SP), CX
 	MOVQ CX, 216(AX)
+	MOVQ R12, 136(AX)
```
klauspost, line 1916: These showed up after re-regenerating.

PR #570, opened 2022-04-25
.github: check code generation

mmcloughlin:
Adds a CI job to verify code generation is up-to-date with the repository.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
+    branches: [ master ]
 
        
 jobs:
@@ -38,7 +39,7 @@ jobs:
         CGO_ENABLED: 1
       run: go test -cpu="1,4" -short -race -v ./...
 
-  gen:
+  generate:
     strategy:
       matrix:
         pkg: [s2, zstd]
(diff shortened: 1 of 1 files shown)


.github/workflows/go.yml
```diff
@@ -38,7 +38,30 @@ jobs:
...
+  generate:
+    strategy:
+      matrix:
+        pkg: [s2, zstd]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17.x
```
mmcloughlin, line 51: Deliberately running this under 1.17 because of the build constraint behavior. Currently `avo` decides how to output them entirely based on the Go version it was compiled with. So under Go 1.18 there are only `go:build` lines, not `+build`. Since the above job checks Go 1.16 I assumed you were trying to keep 1.16 compatibility for now?

https://github.com/mmcloughlin/avo/issues/183#issuecomment-955164201

klauspost: Correct. I keep 3 versions of backcompat.

klauspost:
@mmcloughlin Thanks for the contribution. 👍🏼

PR #572, opened 2022-04-27
zstd: Allow to ignore checksum checking

### LINKED ISSUE #571: zstd: explicitly disable checksum checking
This is follow up of #556. In short: we're decompressing the same files over and over. We measured that checksumming takes ~15% of decompression time. We want to conditionally perform checksumming (likely: for the first time).

Would it be possible to add an option to the decoder, like `WithChecksum(false)`?

WojciechMula:
Closes #571 

There might be scenarios for which checksum validation is not necessary.

### DIFF
--- a/zstd/decoder.go
+++ b/zstd/decoder.go
@@ -439,7 +439,7 @@ func (d *Decoder) nextBlock(blocking bool) (ok bool) {
 		println("got", len(d.current.b), "bytes, error:", d.current.err, "data crc:", tmp)
 	}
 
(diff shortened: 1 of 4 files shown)


klauspost (changes requested):
Are we actually skipping *calculating* the crc anywhere?

You would have to skip `crc.Write(...)` calls, otherwise you are still calculating crc, just not comparing it at the end.

klauspost (approved):
LGTM

PR #580, opened 2022-05-06
zstd: Fix ignored crc

klauspost:
When `IgnoreChecksum(true)` was used, single threaded decodes would not consume CRC.

Bonus: Remove crc build tag for fuzzing.


WojciechMula:
Oh, sorry for that. :( I'll try to add some unit tests to cover that case.

BTW, we already used this feature and in a real workload not checking CRC gave us 1.25x speedup! (Yeah, at cost of data corruption in case of cosmic ray hit.)

klauspost:
No worries.

PR #597, opened 2022-05-19
huff0: remove duplicated code and dead code

WojciechMula:
Method `decompress1X8BitExactly` does exactly what "case 8" in `decompress1X8Bit`,
and is almost a 1-to-1 copy.

### DIFF
--- a/huff0/decompress.go
+++ b/huff0/decompress.go
@@ -307,22 +307,6 @@ func (d *Decoder) Decompress1X(dst, src []byte) ([]byte, error) {
 	bitsLeft := uint8(br.off)*8 + 64 - br.bitsRead
 	for bitsLeft > 0 {
 		br.fill()
-		if false && br.bitsRead >= 32 {
-			if br.off >= 4 {
-				v := br.in[br.off-4:]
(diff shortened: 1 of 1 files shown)


huff0/decompress.go
```diff
@@ -641,90 +622,6 @@ func (d *Decoder) decompress1X8Bit(dst, src []byte) ([]byte, error) {
...
-	bufs := d.buffer()
-	buf := &bufs[0]
-	var off uint8
-
-	const shift = 56
-
-	//fmt.Printf("mask: %b, tl:%d\n", mask, d.actualTableLog)
-	for br.off >= 4 {
-		br.fillFast()
-		v := dt[uint8(br.value>>shift)]
```
klauspost, line 669: Main difference is this being a constant shift in this mode and `br.peekByteFast()` being used below.

I would expect this to be faster in Go.

WojciechMula: Right, I'll benchmark this and get back to that with some results.

WojciechMula: @klauspost you were right, there are significant regressions

Issue #595, opened 2022-05-13
huff0: provide asm implementation of Decompress1X


WojciechMula:
`Decompress1X` uses a similar approach to `Decompress4X` -- it first writes data to an auxiliary buffer, and the buffer is later appended to the output. In an asm we can write directly to the output (see: #577).

WojciechMula:
@klauspost there is dead code: https://github.com/klauspost/compress/blob/master/huff0/decompress.go#L310. Should it be re-enabled?

klauspost:
@WojciechMula Crashers happen after merge :(

Sample [crashers.zip](https://github.com/klauspost/compress/files/8755706/crashers.zip)

Could you PTAL?

klauspost:
Test is rather simple:
```
func FuzzDecompress(data []byte) int {
	s, rem, err := huff0.ReadTable(data, nil)
	if err != nil {
		return 0
	}
	_, err1 := s.Decompress1X(rem)
	_, err4 := s.Decompress4X(rem, huff0.BlockSizeMax-3)
	if err1 != nil && err4 != nil {
		return 0
	}
	return 1
} 
```

klauspost:
```
unexpected fault address 0xffffffffffffffff
fatal error: fault
[signal 0xc0000005 code=0x0 addr=0xffffffffffffffff pc=0xa463ee]

goroutine 1 [running]:
runtime.throw({0xa62546, 0x22466b4ffff})
	runtime/panic.go:1198 +0x76 fp=0xc0000d3c48 sp=0xc0000d3c18 pc=0x9c3196
runtime.sigpanic()
	runtime/signal_windows.go:260 +0x10c fp=0xc0000d3c90 sp=0xc0000d3c48 pc=0x9d46ec
github.com/klauspost/compress/huff0.(*Scratch).compress1X(...)
	e:/gopath/src/github.com/klauspost/compress/huff0/compress.go:230
github.com/klauspost/compress/huff0.(*Scratch).compress1X-fm({0xc0000d3d40, 0x2246c27001b, 0xc0003e0000})
	e:/gopath/src/github.com/klauspost/compress/huff0/compress.go:229 +0x4e fp=0xc0000d3cd8 sp=0xc0000d3c90 pc=0xa463ee
github.com/klauspost/compress/huff0.(*Decoder).Decompress1X(0xc0000d3df0, {0xc0003a0000, 0x0, 0x4000000010}, {0x2246c27001b, 0x100, 0x0})
[...]

PR #604, opened 2022-05-26
huff0: fix handling errors in Decompress1X asm

WojciechMula:
It fixes the regression reported in #595. I didn't add all three input files, as they are quite big. I checked that with the fix all are handled without panics/CPU errors.


klauspost (approved):
LGTM

klauspost:
Running fuzz tests now.

WojciechMula:
> Running fuzz tests now.

Great! Thank you.

PR #627, opened 2022-06-15
s2: Add repeat checks to better mode

klauspost:
Improves compression ~0.1% at ~10% speed cost. Seems too little to keep.

More benchmarking needed.


klauspost:
#629 Adds much more compression for same speed loss.

PR #644, opened 2022-07-17
zstd: Disable extended memory copies (amd64)

klauspost:
Using longer memory copies leads to rare, random crashes with fuzz testing.

It is therefore disabled for now.


WojciechMula:
BTW, do you think it would be easy to overallocate buffers? I mean make sure we have always a few spare bytes to always use the fast copy routine.

PR #649, opened 2022-07-27
zstd: Add DecodeAllCapLimit

### LINKED ISSUE #647: proposal: zstd: add DecodedSize function
I'm trying to implement proper rate-limiting and I'm finding it difficult to adequately rate-limit a server when dealing with untrusted `[]byte` inputs since I don't know how large the decompressed output will end up. I could convert the `[]byte` into a `io.Reader` and perform rate limiting based on the number of bytes read from `zstd.Decoder`, but that seems somewhat unfortunate.

I propose the addition of:
```go
// DecodedLength reports the decompressed length of src in bytes.
func (*Decoder) DecodedLength(src []byte) (int64, error)
```
It may or may not make sense as a method on `Decoder`.

IIUC, this should only need to perform work proportional to O(len(src)). The amount of memory it will occupy is primarily what's needed to initialize the ANS tables and to read each symbol. We would then count each literal symbol and the length of each LZ77 back-reference symbol.

This API also assists in determining what sized buffers to grab for the output.

klauspost:
WithDecodeAllCapLimit will limit DecodeAll to decoding cap(dst)-len(dst) bytes, or any size set in WithDecoderMaxMemory.

This can be used to limit decoding to a specific maximum output size.

Disabled by default.

Fixes #647

### DIFF
--- a/zstd/decoder.go
+++ b/zstd/decoder.go
@@ -502,22 +502,6 @@ func (d *Decoder) nextBlockSync() (ok bool) {
 				d.current.err = ErrDecoderSizeExceeded
 				return false
 			}
-			if d.frame.FrameContentSize != fcsUnknown {
-				if !d.o.limitToCap && d.frame.FrameContentSize > d.o.maxDecodedSize {
-					if debugDecoder {
-						println("decoder size exceeded, fcs:", d.frame.FrameContentSize, "> mds", d.o.maxDecodedSize)
-					}
-					d.current.err = ErrDecoderSizeExceeded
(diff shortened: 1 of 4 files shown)


dsnet (approved):
Thanks! I only did a high-level look through and it generally looks sensible. I'll report if there are any bugs with it.

klauspost:
@dsnet I will merge it, so it is easier for you to test. I will however not release, since I am mostly AFK for the next few weeks, just FYI.

PR #663, opened 2022-09-19
s2: Improve Better compression slightly

klauspost:
Index second part from the middle instead of backwards. Gives small size improvement at the same speed.


klauspost:
I will try indexing forward from the middle.

PR #687, opened 2022-10-31
investigation and proof of concept for plucking files from .zip file

elh:
learning .zip format

I ultimately decided to use `NewReader(r io.ReaderAt, size int64) (*Reader, error)` and solve for my needs using a custom ReaderAt that could use HTTP range requests to grab specific bytes from a remote. That method is not clearly better because you have to do some weird coordination and are fighting against assumptions the lib makes that bytes are local and cheap to read. The optimal performance solution would go down this path of changing the core flow

### DIFF
--- a/test/main.go
+++ b/test/main.go
@@ -1,0 +1,213 @@
+package main
+
+import (
+	"bytes"
+	"compress/flate"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


klauspost:
@elh No problem. Check out: https://github.com/minio/zipindex

elh:
@klauspost nice. super relevant to my problem. thanks, maintainer!

PR #696, opened 2022-11-23
zstd: Faster CRC checking/skipping

greatroar:
Two patches:

* check ignoreChecksum once in Decoder.nextBlock and skip the final bit of computation when it's not needed (restic uses zstd without CRC)
* do the actual check a bit faster and save some allocations by storing the CRC as a uint32 instead of a slice

See second commit for benchmark figures.

### DIFF
--- a/zstd/decoder.go
+++ b/zstd/decoder.go
@@ -459,7 +459,11 @@ func (d *Decoder) nextBlock(blocking bool) (ok bool) {
 		println("got", len(d.current.b), "bytes, error:", d.current.err, "data crc:", tmp)
 	}
 
-	if !d.o.ignoreChecksum && len(next.b) > 0 {
+	if d.o.ignoreChecksum {
+		return true
+	}
+
+	if len(next.b) > 0 {
 		n, err := d.current.crc.Write(next.b)
 		if err == nil {
 			if n != len(next.b) {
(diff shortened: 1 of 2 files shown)


zstd/decoder.go
```diff
@@ -918,18 +919,18 @@ decodeStream:
 				println("next block returned error:", err)
 			}
 			dec.err = err
-			dec.checkCRC = nil
+			dec.hasCRC = false
 			if dec.Last && frame.HasCheckSum && err == nil {
 				crc, err := frame.rawInput.readSmall(4)
-				if err != nil {
+				if len(crc) < 4 {
```
greatroar, line 925: In case it's not obvious, this way of checking for the error triggers BCE in the else branch.

klauspost: No problem, but please make sure that `err` isn't somehow nil. If it is, set it to `io.ErrUnexpectedEOF`.

klauspost (commented):
Just a minor safety check.

zstd/decoder.go
greatroar: Done!

zstd/decoder.go
```diff
@@ -918,18 +919,22 @@ decodeStream:
...
+					if err != nil {
```
klauspost, line 926:
```suggestion
					if err == nil {
```

greatroar: Whoops, fixed.

klauspost (approved):
LGTM

PR #699, opened 2022-11-26
zstd: Improve throughput of SpeedBestCompression encoder

greatroar:
Lifted ofCode and mlCode computations out of match.estBits, so that method will be inlined into its only caller. Also some changes to eliminate a branch: the last if block becomes two CMOVs/CSELs on amd64 and arm64.
    
    name                              old speed      new speed       delta
    Encoder_EncodeAllSimple/best-8    11.1MB/s ± 1%   16.9MB/s ± 1%  +52.23%  (p=0.000 n=10+10)
    Encoder_EncodeAllSimple4K/best-8  8.41MB/s ± 1%  10.95MB/s ± 0%  +30.20%  (p=0.000 n=10+10)
    
    name                              old alloc/op   new alloc/op    delta
    Encoder_EncodeAllSimple/best-8       20.0B ± 0%      18.0B ± 0%  -10.00%  (p=0.002 n=8+10)
    Encoder_EncodeAllSimple4K/best-8     2.00B ± 0%      2.00B ± 0%     ~     (all equal)

### DIFF
--- a/zstd/enc_best.go
+++ b/zstd/enc_best.go
@@ -38,27 +38,29 @@ type match struct {
 const highScore = 25000
 
 // estBits will estimate output bits from predefined tables.
-func (m *match) estBits(bitsPerByte int32) {
-	mlc := mlCode(uint32(m.length - zstdMinMatch))
-	var ofc uint8
-	if m.rep < 0 {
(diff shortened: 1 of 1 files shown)


klauspost (approved):
Beautiful! Thanks!

PR #703, opened 2022-11-30
zstd: Use individual reset threshold

klauspost:
Instead of setting the limit to the largest window size, set it to the chosen one.

Mainly faster on small independent encodes:

```
λ benchcmp before.txt after.txt
benchmark                                         old ns/op     new ns/op     delta
BenchmarkEncoder_EncodeAllSimple4K/fastest-32     3145          3133          -0.38%
BenchmarkEncoder_EncodeAllSimple4K/default-32     41485         40624         -2.08%
BenchmarkEncoder_EncodeAllSimple4K/better-32      49352         49197         -0.31%
BenchmarkEncoder_EncodeAllSimple4K/best-32        421522        407392        -3.35%

benchmark                                         old MB/s     new MB/s     speedup
BenchmarkEncoder_EncodeAllSimple4K/fastest-32     1302.48      1307.39      1.00x
BenchmarkEncoder_EncodeAllSimple4K/default-32     98.74        100.83       1.02x
BenchmarkEncoder_EncodeAllSimple4K/better-32      83.00        83.26        1.00x
BenchmarkEncoder_EncodeAllSimple4K/best-32        9.72         10.05        1.03x
```

### DIFF
--- a/zstd/enc_base.go
+++ b/zstd/enc_base.go
@@ -16,6 +16,7 @@ type fastBase struct {
 	cur int32
 	// maximum offset. Should be at least 2x block size.
 	maxMatchOff int32
+	bufferReset int32
 	hist        []byte
 	crc         *xxhash.Digest
 	tmp         [8]byte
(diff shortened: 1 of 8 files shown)


klauspost:
Stable after 4h fuzz testing.

greatroar:
Nice!

PR #720, opened 2022-12-23
zstd: Shave some instructions off the amd64 asm

greatroar:
These are all in return sequences, so no performance gain expected, but the generated code is smaller.


klauspost (approved):
👍🏼

PR #728, opened 2023-01-05
zstd: Add delta encoding support

greatroar:
This adds support for delta encoding to compress/zstd, compatible with the [--patch-from](https://github.com/facebook/zstd/wiki/Zstandard-as-a-patching-engine) option that was introduced in zstd reference v1.4.5.

The example is a bit simplistic for now.


zstd/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	// Ordinary compression, for reference.
+	w, _ = zstd.NewWriter(nil, bestLevel)
+	z := w.EncodeAll(target, nil)
+	fmt.Printf("compression only: %4d bytes\n", len(z))
+
+	// Output:
+	// delta encoding:     44 bytes
+	// compression only:  111 bytes
```
klauspost, line 46: This will likely change. Tests should not rely on exact compression numbers.

zstd/encoder_options.go
```diff
@@ -316,3 +322,14 @@ func WithEncoderDict(dict []byte) EOption {
...
 	}
 }
+
+// WithEncoderDictRaw registers a dictionary that may be used by the encoder.
+//
+// The slice content may contain arbitrary data. It will be used as an initial
+// history.
+func WithEncoderDictRaw(id uint32, content []byte) EOption {
+	return func(o *encoderOptions) error {
+		o.dict = &dict{id: id, content: content}
```
klauspost, line 332: offsets must be initialized to `[3]uint32{1, 4, 8}`.

klauspost (commented):
Looks good! Some minor things.

zstd/encoder_options.go
greatroar: Done. Also put in a check for dictionaries that are too large for the reference implementation.

zstd/example_test.go
greatroar: I've put in a looser check.

klauspost (approved):
LGTM

zstd/decoder_options.go
```diff
@@ -99,6 +105,15 @@ func WithDecoderDicts(dicts ...[]byte) DOption {
 	}
 }
 
+// WithEncoderDictRaw registers a dictionary that may be used by the decoder.
+// The slice content can be arbitrary data.
+func WithDecoderDictRaw(id uint32, content []byte) DOption {
+	return func(o *decoderOptions) error {
+		o.dicts = append(o.dicts, dict{id: id, content: content})
```
klauspost, line 112: Please add the same as you did to the encoder option.

klauspost (changes requested):

zstd/decoder_options.go
greatroar: Done.

klauspost (approved):

klauspost:
Considered adding a note that the most common data should be at the *end* of the dictionary, since offsets will be smaller and it will compress better. Couldn't really phase it in an easy to understand way, so that can always be added later.

PR #729, opened 2023-01-06
tests: Don't leak file descriptors in tests

greatroar:
I think the GC will clean these up, but you don't want it doing that in the middle of a benchmark run.


klauspost (approved):
👍🏼

PR #765, opened 2023-03-02
ci: set minimal permissions to GitHub Workflows

### LINKED ISSUE #764: CI: Set minimum permissions on GitHub Workflows
Hello!

I'd like to suggest the definition of minimal permissions on your GitHub workflows, as it would harden your security agains supply-chain attacks. I see that one of your workflows ([codeql-analysis](https://github.com/klauspost/compress/blob/master/.github/workflows/codeql-analysis.yml)) declares job-level minimal dependencies, but the other workflows don't actually define them, and they would be set as the GitHub defaults.

The ideal is to always define top-level minimal permissions(that would be inherited by all jobs with unspecified permissions), and then use job-level permissions for any job that might need other specific permissions. In this way you would be secured against erroneous or malicious actions from external jobs you call from your workflow. It's specially important in the case they get compromised, for example.

Setting minimum permissions for workflows is recommended by [GitHub itself](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) and also by other security tools, such as [Scorecards](https://github.com/ossf/scorecard) and [StepSecurity](https://github.com/step-security).

I'd be happy to raise a PR with the changes if you agree.

Context
I'm Diogo and I work on Google's Open Source Security Team([GOSST](https://github.com/diogoteles08#about-gosst-ghost)) in cooperation with the Open Source Security Foundation ([OpenSSF](https://openssf.org/)). My core job is to suggest and implement security changes on widely used open source projects 😊

diogoteles08:
Closes #764

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,6 +20,8 @@ on:
   schedule:
     - cron: '31 1 * * 5'
 
+permissions: read-all
+
 jobs:
   analyze:
     name: Analyze
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,6 +6,7 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions: read-all
 
 jobs:
   build:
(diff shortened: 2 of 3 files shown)


klauspost (approved):
Thanks. LGTM.

PR #767, opened 2023-03-04
gzhttp: Use strings for randomJitter to skip a copy

greatroar:


gzhttp/compress.go
```diff
@@ -236,7 +235,7 @@ func (w *GzipResponseWriter) startGzip(remain []byte) error {
 				}
 				jitRNG = binary.LittleEndian.Uint32(tmp[:])
 			}
-			jit := string(w.randomJitter[:1+jitRNG%uint32(len(w.randomJitter)-1)])
+			jit := w.randomJitter[:1+jitRNG%uint32(len(w.randomJitter)-1)]
 			//fmt.Println("w.buf:", len(w.buf), "remain:", len(remain), "jitter:", len(jit))
 			w.gw.(writer.GzipWriterExt).SetHeader(writer.Header{Comment: &jit})
```
greatroar, line 240: This line still has to copy the string header for jit to the heap. Would it be acceptable for SetHeader to set the Comment field unconditionally? Then it could just be a string instead of a *string.

klauspost:
@greatroar A copy is unavoidable in the end. However this moves it to the implementation: #768

klauspost:
As far as I can tell you are adding more copies with all the string assignments, and making the interface weaker.

I will merge #768 - then you can test and compare directly.

greatroar:
Rebased on master, with the following benchmark results:

    name           old speed      new speed      delta
    2kJitter-8     59.9MB/s ± 4%  60.4MB/s ± 3%  +0.77%  (p=0.035 n=24+25)
    2kJitterRNG-8  79.0MB/s ± 4%  79.5MB/s ± 4%    ~     (p=0.215 n=24+25)
    
    name           old alloc/op   new alloc/op   delta
    2kJitter-8       3.18kB ± 2%    3.19kB ± 2%    ~     (p=0.626 n=23+25)
    2kJitterRNG-8    3.21kB ± 2%    3.21kB ± 3%    ~     (p=0.697 n=25+25)
    
    name           old allocs/op  new allocs/op  delta
    2kJitter-8         16.0 ± 0%      15.0 ± 0%  -6.25%  (p=0.000 n=25+25)
    2kJitterRNG-8      16.0 ± 0%      16.0 ± 0%    ~     (all equal)

klauspost:
Thanks. If you resolve the conflict I'll merge.

greatroar:
Resolved.

gzhttp/compress.go
```diff
@@ -232,7 +231,7 @@ func (w *GzipResponseWriter) startGzip(remain []byte) error {
 						}
 						h.Write(remain)
 					}
-					var tmp [sha256.BlockSize]byte
+					var tmp [sha256.Size]byte
```
greatroar, line 234: Note: changed block size to output size here.

greatroar:
Added an optimization for the CRC computation, see second commit.

klauspost (approved):
LGTM

PR #768, opened 2023-03-06
gzhttp: Remove a few unneeded allocs

klauspost:
```
Before:
Benchmark2kJitter-32    	   55479	     21371 ns/op	  95.83 MB/s	    3575 B/op	      20 allocs/op

After:
Benchmark2kJitter-32    	   54948	     21599 ns/op	  94.82 MB/s	    3451 B/op	      16 allocs/op
```

Don't put too much into the speed. GC is not included. Most of the allocs are in `httptest.ResponseRecorder`.

### DIFF
--- a/gzhttp/compress.go
+++ b/gzhttp/compress.go
@@ -226,7 +226,8 @@ func (w *GzipResponseWriter) startGzip(remain []byte) error {
 					}
 					h.Write(remain)
 				}
-				jitRNG = binary.LittleEndian.Uint32(h.Sum(nil))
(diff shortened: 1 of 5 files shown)


klauspost:
@greatroar

greatroar:
Looks like an improvement, but I don't see how the copy of the jitter would be unavoidable. I've rebased #767 on this PR as a proof of concept. In that version, SetHeader just sets all the fields, which is restrictive, but sufficient to implement HTB.

PR #769, opened 2023-03-07
gzhttp: Use SHA256 as paranoid option

klauspost:
### Paranoid?

The padding size is determined by the remainder of a CRC32 of the content.

Since the payload contains elements unknown to the attacker, there is no reason to believe they can derive any information from this remainder, or predict it.

However, for those that feel uncomfortable with a CRC32 being used for this can enable "paranoid" mode which will use SHA256 for determining the padding.

The hashing itself is about 2 orders of magnitude slower, but in overall terms will maybe only reduce speed by 10%.

Paranoid mode has no effect if buffer is < 0 (non-content aware padding).

2K payload roundtrip difference:

```
Benchmark2kJitter-32    	   67309	     17580 ns/op	 116.50 MB/s	    3478 B/op	      17 allocs/op
Benchmark2kJitterParanoid-32    	   54398	     21564 ns/op	  94.97 MB/s	    3438 B/op	      16 allocs
```

### DIFF
--- a/gzhttp/compress.go
+++ b/gzhttp/compress.go
@@ -8,8 +8,10 @@ import (
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"hash/crc32"
 	"io"
 	"math"
+	"math/bits"
 	"mime"
 	"net"
 	"net/http"
@@ -73,6 +75,7 @@ type GzipResponseWriter struct {
 	setContentType   bool   // Add content type, if missing and detected.
 	suffixETag       string // Suffix to add to ETag header if response is compressed.
 	dropETag         bool   // Drop ETag header if response is compressed (supersedes suffixETag).
+	sha256Jitter     bool   // Use sha256 for jitter.
 	randomJitter     []byte // Add random bytes to output as header field.
 	jitterBuffer     int    // Maximum buffer to accumulate before doing jitter.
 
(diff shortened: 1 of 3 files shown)


greatroar:
Wouldn't xxhash serve just as well? That's ~4x faster, according to [xxhashbench](https://github.com/cespare/xxhash/tree/main/xxhashbench).

klauspost:
@greatroar 

Not here:
```
Benchmark1KSum64-32    	16388875	        72.98 ns/op	14031.83 MB/s
BenchmarkCRC32/poly=Castagnoli/size=1kB/align=0-32      23923491                49.85 ns/op     20543.17 MB/s
```

Benchie similar to the CRC32 one:

```Go
func Benchmark1KSum64(b *testing.B) {
	data := make([]byte, 1024)
	b.SetBytes(1024)

	h := New()
	in := make([]byte, 0, h.Size())

	// Warm up
	h.Reset()
	h.Write(data)
	h.Sum(in)
	in = in[:0]
	for i := 0; i < b.N; i++ {
		h.Reset()
		h.Write(data)
		h.Sum(in)
	}
}
```

gzhttp/compress.go
```diff
@@ -216,18 +221,32 @@ func (w *GzipResponseWriter) startGzip(remain []byte) error {
...
+					// Use only up to "w.jitterBuffer", otherwise the output depends on write sizes.
+					if len(remain) > 0 && len(w.buf) < w.jitterBuffer {
+						remain := remain
+						if len(remain)+len(w.buf) > w.jitterBuffer {
+							remain = remain[:w.jitterBuffer-len(w.buf)]
+						}
+						h.Write(remain)
 					}
-					h.Write(remain)
+					jitRNG = bits.RotateLeft32(h.Sum32(), 19) ^ 0xab0755de
```
greatroar, line 248: What does the rotate + XOR do?

klauspost: Just some random shuffling.

PR #770, opened 2023-03-07
gzhttp: Fix crypto/rand.Read usage

greatroar:
rand.Reader.Read(p) is allowed to return < len(p) bytes and no error, and the Mac implementation sometimes does. I don't know if it will do that for len(p) == 4, but rand.Read is safer in any case.

This is a security fix. I'm sending it in the clear because it only affects an unreleased version of this library.


klauspost (approved):
👍🏼

PR #773, opened 2023-03-10
zstd: Report EOF from byteBuf.readBig

greatroar:
This method was inconsistent with its cousin readerWrapper.readBig in that it returned 0, nil when the input is too short.

I'm not sure what the effects of this inconsistency are, if any. I suppose nearly all suppressed errors would have been reported someplace else.

### DIFF
--- a/zstd/bytebuf.go
+++ b/zstd/bytebuf.go
@@ -54,7 +54,7 @@ func (b *byteBuf) readBig(n int, dst []byte) ([]byte, error) {
 func (b *byteBuf) readByte() (byte, error) {
 	bb := *b
 	if len(bb) < 1 {
-		return 0, nil
+		return 0, io.ErrUnexpectedEOF
 	}
 	r := bb[0]
 	*b = bb[1:]


klauspost (approved):
LGTM

zstd/bytebuf.go
```diff
@@ -54,7 +54,7 @@ func (b *byteBuf) readBig(n int, dst []byte) ([]byte, error) {
 func (b *byteBuf) readByte() (byte, error) {
 	bb := *b
 	if len(bb) < 1 {
-		return 0, nil
+		return 0, io.ErrUnexpectedEOF
```
hitzhangjie, line 57: Please document io.ErrUnExpectedEOF is normal behavior behavior like `dat, err := dec.DecodeAll(...)`.

the `dat` is valid if `len(dat) != 0`, but this kind of manner looks like obey the go convention.

hitzhangjie: well, even `n, err := conn.Write(...)` does like this.
maybe improving the comment is enough.

hitzhangjie: but in my case, `dec.DecodeAll(buf, nil)`, it's valid zstd compressed data, why not return io.EOF instead of io.UnExpectedEOF?

PR #779, opened 2023-03-15
s2: Fix huge block overflow

### LINKED ISSUE #778: s2.Encode a large value panic
I'd wrote a unit test with s2.Encode, found that a panic on such scenario as following
```
v := make([]byte, math.MaxInt32)
if s2.MaxEncodedLen(len(v)) != -1 {
	s2.Encode(nil, v)
}
```
the stack is 
```runtime: g 6: unexpected return pc for github.com/klauspost/compress/s2.encodeBlockAsm called from 0x5000500050005```

### PR DESCRIPTION
**Pull request by klauspost**
Fixes crash on encoding single blocks close to or above 2147483647 bytes on amd64.

Fixes #778

Use unsigned compares in assembly.

Lowers the maximum block size on 32 bit platforms. Realistically you wouldn't be able to get close to it anyway.

### DIFF
--- a/s2/encode.go
+++ b/s2/encode.go
@@ -355,9 +355,12 @@ const inputMargin = 8
 // will be accepted by the encoder.
 const minNonLiteralBlockSize = 32
 
+const intReduction = 2 - (1 << (^uint(0) >> 63)) // 1 (32 bits) or 0 (64 bits)
+
 // MaxBlockSize is the maximum value where MaxEncodedLen will return a valid block size.
 // Blocks this big are highly discouraged, though.
-const MaxBlockSize = math.MaxUint32 - binary.MaxVarintLen32 - 5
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: s2/encode.go [hunk 1]
```diff
@@ -366,7 +369,14 @@ const MaxBlockSize = math.MaxUint32 - binary.MaxVarintLen32 - 5
 // 32 bit platforms will have lower thresholds for rejecting big content.
 func MaxEncodedLen(srcLen int) int {
 	n := uint64(srcLen)
-	if n > 0xffffffff {
+	if intReduction == 1 {
```
greatroar, line 372: Isn't it more obvious to check whether math/bits.UintSize == 32?

klauspost: Ends up with the same. Potato/Potâto

PR #780, opened 2023-03-15
s2: Allow CustomEncoder fallback

klauspost:
If a custom encoder returns a negative value to the compression with regular methods.

+ Split stateful reader+writer code into separate files.


s2/writer.go
```diff
@@ -0,0 +1,1020 @@
...
+		}()
+	}
+	return nil
+}
+
+func (w *Writer) encodeBlock(obuf, uncompressed []byte) int {
+	if w.customEnc != nil {
+		if ret := w.customEnc(obuf, uncompressed); ret >= 0 {
+			return ret
+		}
```
klauspost, line 438: Functional change here.

PR #776, opened 2023-03-12
zstd: Speed up + improve best encoder

greatroar:
```
name                              old speed      new speed      delta
Encoder_EncodeAllSimple/best-8    14.8MB/s ± 3%  20.7MB/s ± 3%   +39.53%  (p=0.000 n=17+19)
Encoder_EncodeAllSimple4K/best-8  11.8MB/s ± 1%  19.2MB/s ± 6%   +62.17%  (p=0.000 n=20+20)

name                              old alloc/op   new alloc/op   delta
Encoder_EncodeAllSimple/best-8       14.0B ± 0%     10.2B ± 8%   -27.07%  (p=0.000 n=16+19)
Encoder_EncodeAllSimple4K/best-8     1.00B ± 0%     0.00B       -100.00%  (p=0.000 n=20+19)
```

Also, compressing enwik9 takes 6.375% less wall clock time.

Output from silesia corpus and enwik9 is about .05% bigger, due to the different order in which comparisons are done:

```
dickens    3222189    3220994 (× 0.99963)
enwik9   259699309  259846164 (× 1.00057)
mozilla   16912341   16912437 (× 1.00001)
mr         3505553    3502823 (× 0.99922)
nci        2289871    2306320 (× 1.00718)
ooffice    2896410    2896907 (× 1.00017)
osdb       3390871    3390548 (× 0.99990)
reymont    1656006    1657380 (× 1.00083)
samba      4326783    4329898 (× 1.00072)
sao        5416932    5416648 (× 0.99995)
webster    9966351    9972808 (× 1.00065)
xml         538378     542277 (× 1.00724)
x-ray      5733061    5733121 (× 1.00001)
total    319554055  319728325 (× 1.00055)
```

This is still smaller than before #705.

PR #783, opened 2023-03-19
zstd: Improve best encoder by extending backwards

greatroar:
In #776, the compression quality of the zstd SpeedBestCompression encoder decreased, although you may have misunderstood that. In any case, here's a patch that improves it quite a bit, at the expense of some of the speed gain from #776. The final output sizes are smaller than they were prior to that PR.

The encoder now extends matches backwards before estimating their encoded size, rather than doing this after selecting the best match based on the partial matches gotten from only looking forward from the initial match offset.

Benchmarks on amd64:

```
name                              old speed      new speed      delta
Encoder_EncodeAllSimple/best-8    20.7MB/s ± 3%  19.0MB/s ± 1%  -8.04%  (p=0.000 n=19+18)
Encoder_EncodeAllSimple4K/best-8  19.2MB/s ± 6%  17.9MB/s ± 1%  -6.86%  (p=0.000 n=20+20)
```

Output sizes on Silesia and enwik9:

```
dickens    3220994    3179697 (× 0.987179)
enwik9   259846164  257481474 (× 0.990900)
mozilla   16912437   16895142 (× 0.998977)
mr         3502823    3473770 (× 0.991706)
nci        2306320    2300580 (× 0.997511)
ooffice    2896907    2888715 (× 0.997172)
osdb       3390548    3368411 (× 0.993471)
reymont    1657380    1639490 (× 0.989206)
samba      4329898    4315020 (× 0.996564)
sao        5416648    5383855 (× 0.993946)
webster    9972808    9887560 (× 0.991452)
xml         542277     541018 (× 0.997678)
x-ray      5733121    5681186 (× 0.990941)
total    319728325  317035918 (× 0.991579)
```

Wall clock time for compressing enwik9 goes up a bit, but is still close to what is was before #776.

### DIFF
--- a/zstd/enc_best.go
+++ b/zstd/enc_best.go
@@ -205,7 +205,22 @@ encodeLoop:
 					panic(fmt.Sprintf("first match mismatch: %v != %v, first: %08x", src[s:s+4], src[offset:offset+4], first))
 				}
 			}
-			cand := match{offset: offset, s: s, length: 4 + e.matchlen(s+4, offset+4, src), rep: rep}
+
+			l := 4 + e.matchlen(s+4, offset+4, src)
+			if rep < 0 {
+				// Extend candidate match backwards as far as possible.
+				tMin := s - e.maxMatchOff
+				if tMin < 0 {
+					tMin = 0
+				}
+				for offset > tMin && s > nextEmit && src[offset-1] == src[s-1] && l < maxMatchLength {
+					s--
+					offset--
+					l++
+				}
+			}
+
+			cand := match{offset: offset, s: s, length: l, rep: rep}
 			cand.estBits(bitsPerByte)
 			if m.est >= highScore || cand.est-m.est+(cand.s-m.s)*bitsPerByte>>10 < 0 {
 				*m = cand
@@ -295,25 +310,10 @@ encodeLoop:
 			s = best.s
 			var seq seq
 			seq.matchLen = uint32(best.length - zstdMinMatch)
-
-			// We might be able to match backwards.
-			// Extend as long as we can.
-			start := best.s
-			// We end the search early, so we don't risk 0 literals
-			// and have to do special offset treatment.
-			startLimit := nextEmit + 1
-
-			tMin := s - e.maxMatchOff
-			if tMin < 0 {
-				tMin = 0
+			if debugAsserts && s <= nextEmit {
+				panic("s <= nextEmit")
 			}
-			repIndex := best.offset
-			for repIndex > tMin && start > startLimit && src[repIndex-1] == src[start-1] && seq.matchLen < maxMatchLength-zstdMinMatch-1 {
-				repIndex--
-				start--
-				seq.matchLen++
-			}
-			addLiterals(&seq, start)
+			addLiterals(&seq, s)
 
 			// rep 0
 			seq.offset = uint32(best.rep)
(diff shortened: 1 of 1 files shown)


zstd/enc_best.go
```diff
@@ -295,25 +310,7 @@ encodeLoop:
 			s = best.s
 			var seq seq
 			seq.matchLen = uint32(best.length - zstdMinMatch)
-
-			// We might be able to match backwards.
-			// Extend as long as we can.
-			start := best.s
-			// We end the search early, so we don't risk 0 literals
-			// and have to do special offset treatment.
-			startLimit := nextEmit + 1
```
greatroar, line 304: This I don't fully understand, because the other left-extending code below did not have this. I omitted it in the new code, and all the tests pass, also with debugAsserts = true.

klauspost: Repeat offsets are slightly different when there are 0 literals.

https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md#repeat-offsets

> There is an exception though, when current sequence's literals_length = 0. In this case, repeated offsets are shifted by one, so an offset_value of 1 means Repeated_Offset2, an offset_value of 2 means Repeated_Offset3, and an offset_value of 3 means Repeated_Offset1 - 1_byte.

This mean if we extend back so there are no literals the repeat values will be wrong or must be adjusted.

klauspost:
I may have checked against an older version prior to v1.15.13 by accident. Either way to big deal if we bring it back.

zstd/enc_best.go
greatroar: Checking against nextEmit+1, which seemed the safest option, increases the output size by 2.6% (!). I've added an assertion instead. Please check if it's in the right place.

klauspost:
Interestingly I see a regression on github-june-2days-2019.json and github-ranks-backup.bin from here: https://files.klauspost.com/compress/

We can rewrite the repeat checks to be:

```Go
		if canRepeat && best.length < goodEnough {
			if s == nextEmit {
				improve(&best, s-offset2, s, uint32(cv), 1)
				improve(&best, s-offset3, s, uint32(cv), 2)
				improve(&best, s-offset1-1, s, uint32(cv), 3)
			}

			// If either no match or a non-repeat match, check at + 1
			if best.rep <= 0 {
				cv32 := uint32(cv >> 8)
				spp := s + 1
				improve(&best, spp-offset1, spp, cv32, 1)
				improve(&best, spp-offset2, spp, cv32, 2)
				improve(&best, spp-offset3, spp, cv32, 3)
				if best.rep < 0 {
					cv32 = uint32(cv >> 24)
					spp += 2
					improve(&best, spp-offset1, spp, cv32, 1)
					improve(&best, spp-offset2, spp, cv32, 2)
					improve(&best, spp-offset3, spp, cv32, 3)
				}
			}
		}
```

You can then remove the entire `// Check offset 2` section and the check right above.

Another minor tweak is `for index0 < s-1 {` -> `for index0 < s {` (2 places).

You are welcome to also set `const goodEnough = 250` - that gives a little as well. Speed is not really affected that much and we want pretty much the best compression we reasonably can, while not wasting CPU.

klauspost:
(sent it a bit early - fixed it above)

zstd/enc_best.go
klauspost: When fully supporting these below, we can remove the check.

klauspost:
Ran you proposal with these changes across the testset and results looks good.

greatroar:
Closing in favor of #784.

PR #793, opened 2023-03-26
zstd: Fix back-referenced offset

klauspost:
Since we expand backwards early, we may be in a situation where best.s+2 has already been indexed.

This will result in picking up a 0 or negative offset, which leads to corrupted data.

Skip this check if best.s is less than or equal to s-2.

Regression from https://github.com/klauspost/compress/pull/784 (not released)


klauspost:
Caught 15 minutes after this PR on https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=57460

PR #794, opened 2023-03-26
zstd: Load source value at start of loop

greatroar:
There were four places were this value was loaded: before the loop, at the end, and before two continues. Doing it at the start reduces the code size. It also seems to give a tiny speedup:

```
name                              old speed      new speed      delta
Encoder_EncodeAllSimple/best-8    16.8MB/s ± 2%  16.9MB/s ± 1%  +0.47%  (p=0.025 n=8+10)
Encoder_EncodeAllSimple4K/best-8  16.0MB/s ± 2%  16.2MB/s ± 1%  +1.67%  (p=0.000 n=10+10)

name                              old alloc/op   new alloc/op   delta
Encoder_EncodeAllSimple/best-8       19.0B ± 5%     18.0B ± 0%  -5.26%  (p=0.002 n=9+9)
Encoder_EncodeAllSimple4K/best-8     1.00B ± 0%     1.00B ± 0%    ~     (all equal)
```

### DIFF
--- a/zstd/enc_best.go
+++ b/zstd/enc_best.go
@@ -159,7 +159,6 @@ func (e *bestFastEncoder) Encode(blk *blockEnc, src []byte) {
 
 	// nextEmit is where in src the next emitLiteral should start from.
 	nextEmit := s
-	cv := load6432(src, s)
 
 	// Relative offsets
 	offset1 := int32(blk.recentOffsets[0])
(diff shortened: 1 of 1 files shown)


greatroar:
I tried doing this for the other encoders as well, but they're structured differently, as nested loops.

klauspost:
The original intention was to reduce latency of the load by doing it as early as possible. When there are less checks this load is very sensitive to latency.

I think the days of this being a worthwhile optimization is gone, especially since there is no condition on this for loop.

klauspost (approved):

PR #795, opened 2023-03-26
zstd: Shorten checksum code

greatroar:
frameDec.consumeCRC is now inlineable. Its callers already check for ignoreChecksum. frameDec.checkCRC's callers already check for HasCheckSum and ignoreChecksum.

xxhash.Digest.Write is documented as always returning len(b), nil, so no need to check its return values.


klauspost (approved):
LGTM

PR #799, opened 2023-04-03
Allow access to internal ResponseWriter in gzhttp handler

jgimenez:
Sometimes users will want to use the `http.Flusher`, or `http.Hijacker`, etc. interfaces from an `http.ResponseWriter`, which `gzhttp.ResponseWriter` does not implement but the underlying one does.

It is common to provide an `Unwrap` function to access it, for example the standard library expects that (although I didn't find any official documentation on it). See [http.ResponseController](https://cs.opensource.google/go/go/+/refs/tags/go1.20.2:src/net/http/responsecontroller.go;l=41-43) for example.

### DIFF
--- a/gzhttp/compress.go
+++ b/gzhttp/compress.go
@@ -169,6 +169,10 @@ func (w *GzipResponseWriter) Write(b []byte) (int, error) {
 	return len(b), nil
 }
 
+func (w *GzipResponseWriter) Unwrap() http.ResponseWriter {
+	return w.ResponseWriter
+}
+
 var castagnoliTable = crc32.MakeTable(crc32.Castagnoli)
 
 // startGzip initializes a GZIP writer and writes the buffer.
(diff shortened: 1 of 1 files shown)


klauspost:
Both [http.Hijacker](https://github.com/klauspost/compress/blob/master/gzhttp/compress.go#L394) and [http.Flusher](https://github.com/klauspost/compress/blob/master/gzhttp/compress.go#L348) are supported and the latter must be used for it to work.

Please add this to [NoGzipResponseWriter](https://github.com/klauspost/compress/blob/master/gzhttp/compress.go#L944) and I will consider whether to include it. Be aware there is [additional wrapping](https://github.com/klauspost/compress/blob/master/gzhttp/compress.go#L927) if the provided ResponseWriter supports hijacking.

klauspost:
(also, you can just check the type and grab it directly - after all it is publicly available).

jgimenez:
Makes sense, done.

My main motivation is for compatibility with `http.ResponseController` from the standard library, linked above, where I can't change how the introspection is done.

klauspost (approved):
LGTM

PR #802, opened 2023-04-07
readerWrapper: readByte needs to use io.ReadFull

jnoxon:
This fixes an "unexpected EOF" error that can occur when streaming from a network source into the zstd decompressor. I maintain a production system which streams zstd-compressed data from AWS S3 that hits this bug regularly when under heavy load.

Even when asking for one byte of data, it's necessary to use io.ReadFull. It's valid to return 0 bytes without EOF on a read, and this behavior is seen on network sources rather frequently.

I wrote up a test that reproduces the bug here:

https://github.com/jnoxon/zstd-zero-length-read-bug


klauspost:
Thanks for the fix!

PR #815, opened 2023-05-02
gzhttp: Handle informational headers

### LINKED ISSUE #814: gziphttp: Support informational headers
Hello!

I'm a maintainer of Traefik, which rely on klauspost/compress for its [Compress middleware](https://doc.traefik.io/traefik/v2.10/middlewares/http/compress/).

We recently updated Traefik to go1.20, which brought the support for 103 Early Hints status code, and more generally 1xx status codes, as per the go1.20 release note:
> The [ResponseWriter.WriteHeader](https://go.dev/pkg/net/http/#ResponseWriter.WriteHeader) function now supports sending 1xx status codes

We gathered feedback on that feature and we identified a bug when using the gzip compression middleware.

Here is a sample curl trace of response with Early Hints, with and without gzip compression:

Without gzip:
```
❯ curl -vvv  -H "Content-Type: text/html" http://whoami:8180
*   Trying 127.0.0.1:8180...
* Connected to whoami (127.0.0.1) port 8180 (#0)
> GET / HTTP/1.1
> Host: whoami:8180
> User-Agent: curl/7.88.1
> Accept: */*
> Content-Type: text/html
> 
< HTTP/1.1 103 Early Hints
< Link: </style.css>; rel=preload; as=style
< Link: </script.js>; rel=preload; as=script
< Vary: Accept-Encoding
< HTTP/1.1 404 Not Found
< Content-Length: 1103
< Content-Type: text/html; charset=utf-8
< Date: Tue, 02 May 2023 08:06:43 GMT
< Link: </style.css>; rel=preload; as=style
< Link: </script.js>; rel=preload; as=script
< Link: </style2.css>; rel=preload; as=style
< Link: </script2.js>; rel=preload; as=script
< 
<!doctype html>
* Connection #0 to host whoami left intact
[... rest of the response body is omitted from the example ...]
```

With gzip:
```
❯ curl -vvv --compressed -H "Content-Type: text/html" http://whoami:8180
*   Trying 127.0.0.1:8180...
* Connected to whoami (127.0.0.1) port 8180 (#0)
> GET / HTTP/1.1
> Host: whoami:8180
> User-Agent: curl/7.88.1
> Accept: */*
> Accept-Encoding: deflate, gzip, br, zstd
> Content-Type: text/html
> 
< HTTP/1.1 103 Early Hints
< Content-Encoding: gzip
< Content-Type: text/html; charset=utf-8
< Date: Tue, 02 May 2023 08:07:07 GMT
< Link: </style.css>; rel=preload; as=style
< Link: </script.js>; rel=preload; as=script
< Link: </style2.css>; rel=preload; as=style
< Link: </script2.js>; rel=preload; as=script
< HTTP/1.1 200 OK
< Content-Encoding: gzip
< Content-Type: text/html; charset=utf-8
< Date: Tue, 02 May 2023 08:07:07 GMT
< Link: </style.css>; rel=preload; as=style
< Link: </script.js>; rel=preload; as=script
< Link: </style2.css>; rel=preload; as=style
< Link: </script2.js>; rel=preload; as=script
< Content-Length: 852
< 
<!doctype html>
* Connection #0 to host whoami left intact
[... rest of the response body is omitted from the example ...]
```

The first problem, that appears in the trace, is the duplication of headers.
The second problem, not shown in the trace, is that the informational headers are transmitted together with the gzip handler's response, so not as an interim response, which defeats the purpose of those headers.

All in all, if we did not overlook something, we think this should be fixed and we are going to open a PR for that.

Edit: I forgot to mention the third problem, which is that the status code is overwritten. The test server we used is replying at last with a `404 Not Found` status code, but the gzip response is a `200 OK` anyway in that case.

rtribotte:
Hello,

This PR introduces the support of 1xx status codes forwarding.
One unit test has been added and is an adapted version of the go standard library test for 1xx status code responses.

Fixes #814


klauspost:
@rtribotte Thanks! I am not too familiar with how 100 responses are handled - or rather why they are special.

I assume the Go library allow to send a 1xx header and then afterwards a 200+ response code? Maybe if you have some information on it.

klauspost (approved):
Very good. Thanks for the details in #814

klauspost:
@rtribotte Seems like you need to gate this for Go 1.20+

rtribotte:
Hey @klauspost,

Thanks for the prompt review!
I pushed a review commit to gate the test for go1.20.

klauspost:
@rtribotte Shouldn't the functionality also be gated?

As far as I can tell, this could break on older Go versions, where if you send a 100, it will be sent upstream, which will send all headers.

For older versions this would make the headers immutable, meaning that Content-Encoding can no longer be set - and the content will be sent compressed, but without the appropriate header.

This is all speculation, so feel free to correct me.

rtribotte:
@klauspost Thanks for the feedback, I think you are right regarding the immutability of headers.

For package `net/http` the 1xx codes are correctly handled starting with v1.19:
go1.18 :https://github.com/golang/go/blob/581603cb7d02019bbf4ff508014038f3120a3dcb/src/net/http/server.go#L1146
go1.19: https://github.com/golang/go/blob/484330d038d060c6e4db3dc8e6ea2b811b2a44d8/src/net/http/server.go#L1168-L1171

For package `x/net/http2` the fix (https://github.com/golang/net/commit/95641704303c734b6cb2959d998ce27a6bf81ae0) is shipped since v0.9.0, which seems to be targeted for go1.17+.

So the specific concern of headers immutability after the first write, I would say that the handling of informational headers should gated to go1.19+.

Nonetheless, for package `net/http/httputil` the fix to forward 1xx responses has been shipped with go1.20: https://github.com/golang/go/commit/972870da11bb4ddd3df393473c494bd2b546cc56

Considering the latter I pushed a new commit to gate the feature for go1.20+.

klauspost (approved):

klauspost:
@rtribotte Thanks. I prefer the conservative choice. Merging when tests have run.

PR #816, opened 2023-05-04
docs: Fix undefined WithDecoderConcurrency

dveeden:
The example would otherwise give a `undefined: WithDecoderConcurrency`


klauspost (approved):
Thanks!

PR #823, opened 2023-06-07
Perf(zstd): Improve 'matchLen' performance by vector instructions.

zzzzwc:
When we use zstd to compress csv text, we find that the matchlen function takes a lot of time, so we try to use vector instructions to speed up the matchlen function.

The following is the benchmark test command and the comparison with the old version:

```
⇒  go test -run=None -bench='Encoder_EncodeAll|Random' -count=6  >  new.txt
⇒  go test -run=None -bench='Encoder_EncodeAll|Random' -count=6  >  old.txt
⇒  benchstat old.txt new.txt
goos: linux
goarch: amd64
pkg: github.com/klauspost/compress/zstd
cpu: Intel(R) Xeon(R) Gold 5218 CPU @ 2.30GHz
                                     │   old.txt    │               new.txt               │
                                     │    sec/op    │    sec/op     vs base               │
Encoder_EncodeAllXML-64                19.13m ±  3%   18.13m ±  2%   -5.21% (p=0.002 n=6)
Encoder_EncodeAllSimple/fastest-64     328.7µ ±  3%   344.9µ ±  2%   +4.94% (p=0.002 n=6)
Encoder_EncodeAllSimple/default-64     489.9µ ±  4%   500.9µ ±  3%   +2.24% (p=0.026 n=6)
Encoder_EncodeAllSimple/better-64      634.4µ ±  3%   632.9µ ±  3%        ~ (p=0.937 n=6)
Encoder_EncodeAllSimple/best-64        2.674m ±  6%   2.680m ±  1%        ~ (p=0.937 n=6)
Encoder_EncodeAllSimple4K/fastest-64   16.84µ ±  3%   16.99µ ±  8%        ~ (p=0.818 n=6)
Encoder_EncodeAllSimple4K/default-64   44.17µ ±  3%   46.14µ ±  1%   +4.45% (p=0.002 n=6)
Encoder_EncodeAllSimple4K/better-64    56.05µ ±  3%   55.94µ ±  3%        ~ (p=0.732 n=6)
Encoder_EncodeAllSimple4K/best-64      285.4µ ± 25%   292.3µ ±  3%        ~ (p=0.485 n=6)
Encoder_EncodeAllHTML-64               315.6µ ±  1%   308.4µ ±  3%   -2.27% (p=0.002 n=6)
Encoder_EncodeAllTwain-64              4.579m ±  1%   4.717m ±  3%   +3.02% (p=0.015 n=6)
[...]

PR #824, opened 2023-06-09
zstd: Add amd64 match length assembly

klauspost:
Copied from the S2 implementation. 5-10% faster.

(Potentially) Replaces #823


klauspost:
(fuzz tested 3 hours)

PR #832, opened 2023-06-22
s2: add GetBufferCapacity() method

GiedriusS:
Add GetBufferCapacity() method. We are reusing readers with sync.Pool and we'd like to avoid allocating memory for the default block size since most of the inputs are smaller. To have a better estimate of how big the lazy buffer should be, we are thinking about keeping in mind a running average of the internal buffer capacities. This method would allow us to implement that.

### DIFF
--- a/s2/reader.go
+++ b/s2/reader.go
@@ -147,6 +147,13 @@ type Reader struct {
 	ignoreCRC      bool
 }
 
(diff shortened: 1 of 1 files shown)


klauspost:
Sounds reasonable.

PR #835, opened 2023-07-10
chore: use copy instead of a loop

testwill:


klauspost:
Copies are overlapping. Therefore your code is not equivalent.

PR #851, opened 2023-08-10
tests: Use Go 1.21.x

klauspost:


klauspost:
ref: https://github.com/burrowers/garble/issues/784

klauspost:
ref: https://github.com/burrowers/garble/issues/785

PR #860, opened 2023-09-16
gzip: Copy bufio.Reader on Reset

thatguystone:
The code already checks to see if the buffer can be reused, but since it's not copied in the overwrite, a new buffer is allocated each time.


klauspost (approved):
Thanks!

PR #864, opened 2023-09-25
zstd: Reject empty compressed blocks in streams

klauspost:
#862 (unreleased) only rejected in-memory encodes - reject for streams as well.

Fixes https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=62528


klauspost:
Requested spec clarification: https://github.com/facebook/zstd/issues/3775

klauspost:
This was a problem with older zstd versions. Reverting #862 instead.

PR #869, opened 2023-10-09
Reduce allocations in flate decompressor and minor code improvements

fakefloordiv:
- `decompressor.step` is now enumeric instead of function pointer. This reduces ALL allocations in decompressor benchmarks to constant 1 (previously this number differed for each case). However, it doesn't bring much gains in terms of throughtput.
- `huffmanBlockDecoder()`, generated by _gen/gen_inflate.go, also used to return a function pointer. Now calling the method directly inside of it. Also doesn't bring much gains, but in microbenchmarks may be visible
- Improved decompressor benchmarks by replacing io.Copy() with io.CopyBuffer(), because io.Copy() allocates internally a buffer for reading
- Removed some expressions, that are useless

### DIFF
--- a/flate/inflate.go
+++ b/flate/inflate.go
@@ -120,8 +120,9 @@ func (h *huffmanDecoder) init(lengths []int) bool {
 	const sanity = false
 
 	if h.chunks == nil {
-		h.chunks = &[huffmanNumChunks]uint16{}
+		h.chunks = new([huffmanNumChunks]uint16)
 	}
+
 	if h.maxRead != 0 {
 		*h = huffmanDecoder{chunks: h.chunks, links: h.links}
 	}
@@ -352,7 +353,7 @@ func (f *decompressor) nextBlock() {
 		// compressed, fixed Huffman tables
 		f.hl = &fixedHuffmanDecoder
 		f.hd = nil
-		f.huffmanBlockDecoder()()
+		f.huffmanBlockDecoder()
 		if debugDecode {
 			fmt.Println("predefinied huffman block")
 		}
@@ -363,7 +364,7 @@ func (f *decompressor) nextBlock() {
 		}
 		f.hl = &f.h1
 		f.hd = &f.h2
-		f.huffmanBlockDecoder()()
+		f.huffmanBlockDecoder()
 		if debugDecode {
 			fmt.Println("dynamic huffman block")
 		}
--- a/flate/inflate.go
+++ b/flate/inflate.go
@@ -675,7 +675,6 @@ func (f *decompressor) copyData() {
 	if f.dict.availWrite() == 0 || f.copyLen > 0 {
 		f.toRead = f.dict.readFlush()
 		f.step = copyData
-		//f.step = (*decompressor).copyData
 		return
 	}
 	f.finishBlock()
@@ -689,7 +688,6 @@ func (f *decompressor) finishBlock() {
 		f.err = io.EOF
 	}
 	f.step = nextBlock
-	//f.step = (*decompressor).nextBlock
 }
 
 // noEOF returns err, unless err == io.EOF, in which case it returns io.ErrUnexpectedEOF.
--- a/flate/inflate.go
+++ b/flate/inflate.go
@@ -201,8 +201,7 @@ func (h *huffmanDecoder) init(lengths []int) bool {
 			if cap(h.links[off]) < numLinks {
 				h.links[off] = make([]uint16, numLinks)
 			} else {
-				links := h.links[off][:0]
-				h.links[off] = links[:numLinks]
+				h.links[off] = h.links[off][:numLinks]
 			}
 		}
 	} else {
--- a/flate/inflate.go
+++ b/flate/inflate.go
@@ -43,6 +43,9 @@ var bitMask32 = [32]uint32{
 	0x1ffFFFF, 0x3ffFFFF, 0x7ffFFFF, 0xfffFFFF, 0x1fffFFFF, 0x3fffFFFF, 0x7fffFFFF,
 } // up to 32 bits
 
+// zeroChunks is used to nullify decompressor.chunks array
+var zeroChunks = make([]uint16, huffmanNumChunks)
+
 // Initialize the fixedHuffmanDecoder only once upon first use.
 var fixedOnce sync.Once
 var fixedHuffmanDecoder huffmanDecoder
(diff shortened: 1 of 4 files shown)


flate/inflate.go
```diff
@@ -175,9 +176,8 @@ func (h *huffmanDecoder) init(lengths []int) bool {
 	}
 
 	h.maxRead = min
-	chunks := h.chunks[:]
-	for i := range chunks {
-		chunks[i] = 0
+	for i := range h.chunks {
```
klauspost, line 179: This still copies "h.chunks" AFAIK. Arrays should be sliced for using in for.

flate/inflate.go
```diff
@@ -379,7 +390,24 @@ func (f *decompressor) Read(b []byte) (int, error) {
 		if f.err != nil {
 			return 0, f.err
 		}
-		f.step(f)
+
+		switch f.step {
```
klauspost, line 394: Let's put this into a `func (f *decompressor) execStep()` to avoid the duplication.

klauspost (changes requested):
Thanks!

Could you post times in a (comparable) benchmark? Just want to ensure there isn't a speed regression.

fakefloordiv:
@klauspost in `doStep()` method, if current enum is unknown - nothing will be done. This may result in undefined behaviour, up to infinite loop. Currently there's just a debug print in default case, however this doesn't help much either. Shall we rise a panic there?

flate/inflate.go
```diff
@@ -175,10 +179,10 @@ func (h *huffmanDecoder) init(lengths []int) bool {
...
 
 	h.maxRead = min
-	chunks := h.chunks[:]
-	for i := range chunks {
-		chunks[i] = 0
-	}
+
+	// instead of iterating over the whole array, just copy already null-filled
+	// slice in it.
+	copy(h.chunks[:], zeroChunks)
```
klauspost, line 185: Copying is pointless and just slower. This loop is already being picked up by the compiler.

If you want a neat oneliner use `*h.chunks = [huffmanNumChunks]uint16{}`. It produces the exact same code as the current (using DUFFZERO).

klauspost:
> if current enum is unknown - nothing will be done. This may result in undefined behaviour, up to infinite loop. Currently there's just a debug print in default case, however this doesn't help much either. Shall we rise a panic there?

Sure. Should be easier to pick up by a Fuzzer then.

flate/inflate.go
fakefloordiv: I did some benchmarks, and original approach is the most performant (0.3ns faster). Copying and zeroing via for-loop without slicing are at the same level of 8.5ns, and reassigning is ~2x slower (14ns). Sorry for ignorance, I'm gonna return everything back just as it was:)

fakefloordiv:
Full benchmarks output can be found here: https://gist.github.com/fakefloordiv/395a15a982cda3e43dc4f4833d3b2aac

Here's short conclusion:

Time per operation (in ns/op):
```
DecodeDigitsSpeed1e4:   36650   -> 33963 (1.079x)
DecodeDigitsSpeed1e5:   346724  -> 340885 (1.017x)
DecodeDigitsSpeed1e6:   3576404 -> 3515543 (1.017x)
DecodeDigitsDefault1e4:   32550   -> 31307 (1.04x)
DecodeDigitsDefault1e5:   388516  -> 383745 (1.012x)
DecodeDigitsDefault1e6:   3847923 -> 3825121 (1.006x)
DecodeDigitsCompress1e4:   33756   -> 31259 (1.08x)
DecodeDigitsCompress1e5:   367638  -> 359207 (1.023x)
DecodeDigitsCompress1e6:   3618779 -> 3546951 (1.02x)
DecodeTwainSpeed1e4:   28407   -> 27533 (1.032x)
DecodeTwainSpeed1e5:   377619  -> 371688 (1.016x)
DecodeTwainSpeed1e6:   3810692 -> 3726031 (1.023x)
DecodeTwainDefault1e4:   29461   -> 27603 (1.067x)
DecodeTwainDefault1e5:   393549  -> 385159 (1.022x)
DecodeTwainDefault1e6:   3926477 -> 3854784 (1.019x)
DecodeTwainCompress1e4:   26872   -> 25846 (1.04x)
DecodeTwainCompress1e5:   365831  -> 358215 (1.021x)
DecodeTwainCompress1e6:   3651813 -> 3577634 (1.021x)
DecodeRandomSpeed1e4:   214     -> 202.8 (1.055x)
DecodeRandomSpeed1e5:   1659    -> 1626 (1.02x)
DecodeRandomSpeed1e6:   17198   -> 17274 (0.996x)
```

Throughtput (in mb/s):
```
DecodeDigitsSpeed1e4:   272.85 -> 294.43 (1.079x)
DecodeDigitsSpeed1e5:   288.41 -> 293.35 (1.017x)
DecodeDigitsSpeed1e6:   279.61 -> 284.45 (1.017x)
DecodeDigitsDefault1e4:   307.22 -> 319.42 (1.04x)
DecodeDigitsDefault1e5:   257.39 -> 260.59 (1.012x)
DecodeDigitsDefault1e6:   259.88 -> 261.43 (1.006x)
DecodeDigitsCompress1e4:   296.24 -> 319.91 (1.08x)
DecodeDigitsCompress1e5:   272.01 -> 278.39 (1.023x)
DecodeDigitsCompress1e6:   276.34 -> 281.93 (1.02x)
DecodeTwainSpeed1e4:   352.03 -> 363.2 (1.032x)
DecodeTwainSpeed1e5:   264.82 -> 269.04 (1.016x)
DecodeTwainSpeed1e6:   262.42 -> 268.38 (1.023x)
DecodeTwainDefault1e4:   339.44 -> 362.28 (1.067x)
DecodeTwainDefault1e5:   254.1 -> 259.63 (1.022x)
DecodeTwainDefault1e6:   254.68 -> 259.42 (1.019x)
DecodeTwainCompress1e4:   372.13 -> 386.91 (1.04x)
DecodeTwainCompress1e5:   273.35 -> 279.16 (1.021x)
DecodeTwainCompress1e6:   273.84 -> 279.51 (1.021x)
DecodeRandomSpeed1e4:   46695.1 -> 49300.24 (1.056x)
DecodeRandomSpeed1e5:   60272.24 -> 61516.82 (1.021x)
DecodeRandomSpeed1e6:   58144.68 -> 57889.09 (0.996x)
```

klauspost:
> benchmarks were also updated

Yes. That is why I asked for "comparable benchmarks", meaning the same code. :)

I don't think the "CopyBuffer" will make any difference, since both will use the `io.WriterTo` interface, so you should be able to take that out.

fakefloordiv:
> > benchmarks were also updated
> 
> Yes. That is why I asked for "comparable benchmarks", meaning the same code. :)
> 
> I don't think the "CopyBuffer" will make any difference, since both will use the `io.WriterTo` interface, so you should be able to take that out.

You are right, there's completely no difference. Here's result with io.Copy: https://gist.github.com/fakefloordiv/d32971935516f566006f28c6c000a330

klauspost (approved):
LGTM

PR #876, opened 2023-10-22
zstd: Fix corrupted output in "best"

### LINKED ISSUE #875: zstd: Data corruption on SpeedBestCompression level
Following a bug report at https://github.com/restic/restic/issues/4523, I've traced the data corruption back to the zstd library.

The following two minimized data chunks return corrupt data after compressing it at SpeedBestCompression level and decompressing it afterwards. There is no data corruption at the default compression level. I've tested versions v1.16.7 (via restic using Go 1.20.7) and v1.17.1 (using Go 1.21.3).

[correct-3d0e366bad4a5e9b443f407b114756a0f5a8153dbc242e0b2601c707136815eb.bin-minimized.txt](https://github.com/klauspost/compress/files/13061692/correct-3d0e366bad4a5e9b443f407b114756a0f5a8153dbc242e0b2601c707136815eb.bin-minimized.txt)
[correct-dc82d97be7683ecd41097ab02d7b15de81e8bbcd1c476c50b254b1f458090929.bin-minimized.txt](https://github.com/klauspost/compress/files/13061693/correct-dc82d97be7683ecd41097ab02d7b15de81e8bbcd1c476c50b254b1f458090929.bin-minimized.txt)

To produce the minimized examples, I've used the following code snippet, which also serves as a reproducer:
```go
package main

import (
	"bytes"
	"os"
	"sort"

)

func buildEncoder() *zstd.Encoder {
	level := zstd.SpeedBestCompression

	opts := []zstd.EOption{
		// Set the compression level configured.
		zstd.WithEncoderLevel(level),
		// Disable CRC, we have enough checks in place, makes the
		// compressed data four bytes shorter.
		zstd.WithEncoderCRC(false),
		// Set a window of 512kbyte, so we have good lookbehind for usual
		// blob sizes.
		zstd.WithWindowSize(512 * 1024),
	}

	enc, err := zstd.NewWriter(nil, opts...)
	if err != nil {
		panic(err)
	}
	return enc
}

func buildDecoder() *zstd.Decoder {
	opts := []zstd.DOption{
		// Use all available cores.
		zstd.WithDecoderConcurrency(0),
		// Limit the maximum decompressed memory. Set to a very high,
		// conservative value.
		zstd.WithDecoderMaxMemory(16 * 1024 * 1024 * 1024),
	}

	dec, err := zstd.NewReader(nil, opts...)
	if err != nil {
		panic(err)
	}
	return dec
}

var enc = buildEncoder()
var dec = buildDecoder()

func verifyCompression(data []byte) bool {
	compressed := enc.EncodeAll(data, nil)
	decompressed, err := dec.DecodeAll(compressed, make([]byte, 0, len(data)))
	if err != nil {
		panic(err)
	}

	return bytes.Equal(data, decompressed)
}

func main() {
	data, err := os.ReadFile(os.Args[1])
	if err != nil {
		panic(err)
	}

	idx := sort.Search(len(data), func(i int) bool {
		return !verifyCompression(data[:i])
	})
	if verifyCompression(data[:idx]) {
		panic("missing compression error")
	}

	startIdx := sort.Search(idx, func(i int) bool {
		return verifyCompression(data[i:idx])
	}) - 1
	if verifyCompression(data[startIdx:idx]) {
		panic("missing compression error")
	}

	println("minimal example from", startIdx, "to", idx)

	f, err := os.OpenFile(os.Args[1]+"-minimized", os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0666)
	if err != nil {
		panic(err)
	}
	_, err = f.Write(data[startIdx:idx])
	if err != nil {
		panic(err)
	}
	err = f.Close()
	if err != nil {
		panic(err)
	}
}
```

Run it using 
```bash
go mod init test
go mod tidy
go run main.go correct-3d0e366bad4a5e9b443f407b114756a0f5a8153dbc242e0b2601c707136815eb.bin
```

klauspost:
Regression from #784 and followup #793

Fixes #875

A 0 offset backreference was possible when "improve" was successful twice in a row in the "skipBeginning" part, only finding 2 (previously unmatched) length 4 matches, but where start offset decreased by 2 in both cases.

This would result in output where the end offset would equal to the next 's', thereby doing a self-reference.

Add a general check in "improve" and just reject these. Will also guard against similar issues in the future.

This also hints at some potentially suboptimal hash indexing - but I will take that improvement separately.

Fuzz test set updated.


zstd/enc_best.go
```diff
@@ -197,12 +197,13 @@ encodeLoop:
...
 		improve := func(m *match, offset int32, s int32, first uint32, rep int32) {
-			if s-offset >= e.maxMatchOff || load3232(src, offset) != first {
+			delta := s - offset
+			if delta >= e.maxMatchOff || delta <= 0 || load3232(src, offset) != first {
 				return
 			}
 			if debugAsserts {
-				if offset <= 0 {
-					panic(offset)
+				if offset >= s {
```
MichaelEischer, line 205: That debug check can never be triggered (`offset >= s` <=> `0 >= offset -s` <=> `0 >= delta`) now that the previous `if [...] delta <= 0` has been added. Is that intentional?

zstd/enc_best.go
```diff
@@ -343,7 +344,7 @@ encodeLoop:
 		if best.rep > 0 {
 			var seq seq
 			seq.matchLen = uint32(best.length - zstdMinMatch)
-			if debugAsserts && s <= nextEmit {
+			if debugAsserts && s < nextEmit {
 				panic("s <= nextEmit")
```
MichaelEischer, line 348: The panic message, probably should also be updated?

zstd/enc_best.go
klauspost: I don't mind debugAsserts being a bit "excessive" in its testing. It is removed from regular builds anyway, and mostly for debugging.

zstd/enc_best.go
klauspost: Agree

PR #885, opened 2023-11-15
zstd: Reject blocks where reserved values are not 0

klauspost:
ref: https://github.com/facebook/zstd/issues/3821


klauspost:
Upstream was merged. We update as well.

PR #886, opened 2023-11-18
huff0: Remove byteReader

greatroar:
There's a byteReader in huff0, probably copied from fse, that is hardly used. The only method that is used is remain, but that just returns the input length as long as no other methods are called. This PR replaces it with a stored input length field on huff0.Scratch.


klauspost (approved):
Great!

PR #893, opened 2023-12-06
huff0: Improve fuzz base set

klauspost:


lizthegrey:
Picked this up — most of it had already landed independently (`internal/fuzz.ReturnFromZip` is on master and used by `s2/fuzz_test.go`'s decoder fuzzers). The part that hadn't shipped — seeding `FuzzDecompress1x` with real compressed output and exercising `Decompress4X`, plus seeding `regression.zip` into `FuzzCompress` — is rebased onto current master in #1208.

Fine to close this one in favor of that if it looks right to you.

PR #899, opened 2023-12-15
CI: Hash pin sensitive actions and configure Dependabot to automatically update them

### LINKED ISSUE #898: CI: Suggest hash-pinning sensitive workflow dependencies
Hi, I'm Diogo and I'm back (see #764 and #809). At the issue #764 I suggested you to set minimal permissions to your workflows. Now I'm coming back to suggest a modification that would provide extra safety for the workflows that yet require dangerous permissions (e.g., contents: write or repository-projects: write).

### Problem

Your workflows [release.yml](https://github.com/klauspost/compress/blob/master/.github/workflows/release.yml) and [codeql-analysis.yml](https://github.com/klauspost/compress/blob/master/.github/workflows/codeql-analysis.yml) are using dangerous permissions and secrets while running external dependencies pinned only by tag. Those patterns could be dangerous because if any of those actions get hijacked (and at the end they're all repositories and are susceptible to attacks like any other), an attacker could change the code that your tags point to, gaining access to your secrets and/or write permissions to your repository.

### Proposed Solution

A simple solution for this problem would be to hash-pin those sensitive actions, pointing the actions to the very specific commit of that release. It follows an example of the change:

```
- uses: r-lib/actions/pr-fetch@v1 
would become
- uses: r-lib/actions/pr-fetch@11a22a908006c25fe054c4ef0ac0436b1de3edbe # v1.3.1
```

And this would enforce that your action is always running at the expected code.

The only downsize of this solution is that it gets trickier to manually update the version of the actions as they get out-of-date, but that can be solved by using a Dependency-Update-Tool (like dependabot or renovatebot). We can configure it to update all workflow dependencies in a single monthly PR, for example -- for security updates, the tool would immediately create the PRs regardless of the frequency set. For the case of the hash-pinning, the PRs would still keep a comment with the human-readable version used =).

### Conclusion

I'll take the liberty of raising a PR implementing my suggestions, so that it becomes easier for you to evaluate. Let me know if you have any questions or concerns.

Thanks!

diogoteles08:
Closes #898

As mentioned on the issue, this PR enhances project security by hash-pinning the dependencies that are called under dangerous permissions. Additionally, it enables dependabot to update them automatically.

I configured dependabot in a way that all of version updates will be collapsed in a single PR sent monthly -- this avoids noisy PRs, which is a common concern haha. Regardless of the frequency chosen, for the case of security updates a PR with the fixed version would be sent right away.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,11 +40,11 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3.6.0
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@03e7845b7bfcd5e7fb63d1ae8c61b0e791134fab # v2.22.11
       with:
         languages: ${{ matrix.language }}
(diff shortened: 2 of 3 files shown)


klauspost (approved):
LGTM

klauspost:
Thanks!

PR #903, opened 2023-12-19
internal/race,s2: add some race instrumentation

egonelbre:
Assembly code isn't instrumented with race detector. This adds few annotations for race detector.


klauspost (approved):
I will fix up the executable `go.mod` files. Not really sure why it overrules this.

PR #913, opened 2024-01-24
zstd: Limit better/best default window to 8MB

klauspost:
Limit the default window size to 8MB on Better/Best.


klauspost:
@jameshartig Didn't realize Chrome had zstd :D

PR #916, opened 2024-01-25
s2: Fix callbacks for skippable blocks and disallow 0xfe (Padding) for custom use

Jille:
The spec writes that 0xfe must not be interpreted by decompressors, so setting up a callback should be denied.

I left 0xfe open for use by the Writer, as it could maybe be valid for a user to handle padding manually.


klauspost:
I guess whoever requested it never tested it...  oh well...

klauspost:
We can always lift the restriction later if needed.

PR #927, opened 2024-02-09
s2: Add AsyncFlush method: Complete the block without flushing

Jille:
My use case is to transfer a large compressed S2 stream with a few changes very often. To get a small diff I want to end blocks at application decided points rather than at byte offsets. This allows me to remove the first byte without every single block changing.

Flush() works for this, but it limits concurrency because it waits for the last block to be compressed rather than allowing that asynchronously.

So I'd like to propose AsyncFlush, which flushes the buffer to a block, but doesn't flush the block to the io.Writer.

There were actually a few places in the s2 code that also wanted to end the block, but didn't necessary want to flush to the writer.

### DIFF
--- a/s2/writer.go
+++ b/s2/writer.go
@@ -716,8 +716,8 @@ func (w *Writer) writeSync(p []byte) (nRet int, errRet error) {
 	return nRet, nil
 }
 
-// AsyncFlush writes any buffered bytes to a block, but does not flush it to the
-// underlying io.Writer.
+// AsyncFlush writes any buffered bytes to a block and starts compressing it.
+// It does not wait for the output has been written as Flush() does.
 func (w *Writer) AsyncFlush() error {
 	if err := w.err(nil); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


s2/writer.go
```diff
@@ -716,9 +716,9 @@ func (w *Writer) writeSync(p []byte) (nRet int, errRet error) {
 	return nRet, nil
 }
 
-// Flush flushes the Writer to its underlying io.Writer.
-// This does not apply padding.
-func (w *Writer) Flush() error {
+// AsyncFlush writes any buffered bytes to a block, but does not flush it to the
+// underlying io.Writer.
```
klauspost, line 720:
```suggestion
// AsyncFlush writes any buffered bytes to a block and starts compressing it.
// It does not wait for the output has been written as Flush() does.
```

klauspost (commented):
Wonderful change. The only thing I would tweak slightly is the documentation.

Jille:
Sounds good to me! Thanks :)

klauspost (approved):
Lgtm

PR #977, opened 2024-07-01
Fix gzhttp transport response.

DeyiXu:
1. Fix response header Content-Encoding: gzip,zstd but not decompress
2. Server side Content-Encoding is not controlled.
Fix the decompression problem when Content-Encoding appears gzip and zstd at the same time

### DIFF
--- a/gzhttp/transport.go
+++ b/gzhttp/transport.go
@@ -69,7 +69,7 @@ type gzRoundtripper struct {
 }
 
 func (g *gzRoundtripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	var requestedComp bool
+
 	if req.Header.Get("Accept-Encoding") == "" &&
 		req.Header.Get("Range") == "" &&
 		req.Method != "HEAD" {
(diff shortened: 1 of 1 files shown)


DeyiXu:
I am implementing a caching middleware, and during use I found that when saving files locally, gzip data that could not be decompressed appeared.

If the data returned through the gzhttp transport middleware needs to be processed and decompressed gzip a second time.

If it can't solve the data perfectly, then what's the point of this middleware's existence?

DeyiXu:
When accessing an uncertain backend HTTP Server, there is no guarantee that the backend server will fully comply with HTTP standards. There may be some data that does not meet the standards. At this time, the data needs to be processed.

The PR I submitted is to solve this problem.

If there is a problem with my code, please reject my PR and I will continue to improve it.

DeyiXu:
Thanks.

I think the option is feasible.

DeyiXu:
Looking forward to your code.

klauspost:
See #978

PR #982, opened 2024-07-26
s2: Add EncodeBuffer buffer recycling callback

### LINKED ISSUE #981: s2: Proposal: Callback when done with EncodeBuffer
Buffers given to s2.Writer.EncodeBuffer() can't be touched until a full Flush/Close is done, while the buffer is probably reusable much sooner.

What would you think of a WriterOption that installs a callback that is invoked whenever it is safe to reuse the buffer given to EncodeBuffer?

I'm thinking `func WriterEncodeBufferCallback(callback func([]byte)) WriterOption`.

The common use case would be to return the buffer to a sync.Pool. I want to pass data straight from a ringbuffer-of-bytes into EncodeBuffer and want to know when I can mark it as freed.

I'll be happy to write up the PR if you agree with this proposal.

klauspost:
WriterBufferDone will perform a callback when EncodeBuffer has finished writing a buffer to the output and the buffer can safely be reused.

If the buffer was split into several blocks, it will be sent after the last block.

Callbacks will not be done concurrently.

Fixes #981

### DIFF
--- a/s2/writer.go
+++ b/s2/writer.go
@@ -83,11 +83,14 @@ type Writer struct {
 	snappy            bool
 	flushOnWrite      bool
 	appendIndex       bool
+	bufferCB          func([]byte)
 	level             uint8
 }
 
 type result struct {
 	b []byte
+	// return when writing
+	ret []byte
 	// Uncompressed start offset
 	startOffset int64
 }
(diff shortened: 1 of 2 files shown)


Jille:
Took me a bit to realize this can't be simplified by freeing it after use immediately because of the block-splitting case.

LGTM!

klauspost:
Yeah. The point of `EncodeBuffer` is to avoid having to copy the input unlike `Write`, so we need to keep it until all blocks have finished. Luckily the synchronization is trivial, since we have to write in order.

PR #986, opened 2024-08-11
Fix typos

deining:
This PR fixes a various typos I spotted in the project.

### DIFF
--- a/compressible_test.go
+++ b/compressible_test.go
@@ -8,7 +8,7 @@ import (
 
 func BenchmarkEstimate(b *testing.B) {
 	b.ReportAllocs()
-	// (predictable, low entropy distibution)
+	// (predictable, low entropy distribution)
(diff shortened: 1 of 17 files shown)


compressible_test.go
```diff
@@ -45,7 +45,7 @@ func BenchmarkEstimate(b *testing.B) {
 		b.Log(Estimate(testData))
 	})
 
-	// (not predictable, high entropy distibution)
+	// (not predictable, high entropy length)
```
klauspost, line 48: You started replacing with length here for some reason.

klauspost (changes requested):
A couple of non-sensical replacements.

compressible_test.go
deining: > You started replacing with length here for some reason.

Sorry, that wasn't on purpose. This is fixed now.

klauspost (approved):

PR #1002, opened 2024-09-10
[gzhttp] Add supported decompress request body

mirecl:
Add new feature from discussions https://github.com/klauspost/compress/discussions/1001

### DIFF
--- a/gzhttp/compress.go
+++ b/gzhttp/compress.go
@@ -464,11 +464,29 @@ func NewWrapper(opts ...option) (func(http.Handler) http.HandlerFunc, error) {
 	return func(h http.Handler) http.HandlerFunc {
 		return func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Add(vary, acceptEncoding)
-			if contentGzip(r) {
+			if c.allowCompressedRequests && contentGzip(r) {
 				r.Header.Del(contentEncoding)
 
-				if readerGzipBody, err := gzip.NewReader(r.Body); err == nil {
-					r.Body = io.NopCloser(readerGzipBody)
+				var (
+					err error
+					gz  *gzip.Reader
+				)
+
+				zr, ok := gzReaderPool.Get().(*gzip.Reader)
+				if ok {
+					gz, err = zr, zr.Reset(r.Body)
+				} else {
+					gz, err = gzip.NewReader(r.Body)
+				}
+
+				defer func() {
+					gzReaderPool.Put(gz)
+					gz = nil
+				}()
+
+				if err == nil {
+					r.Body = io.NopCloser(gz)
+					defer r.Body.Close()
 				}
 			}
 
@@ -544,17 +562,18 @@ func (pct parsedContentType) equals(mediaType string, params map[string]string) 
 
 // Used for functional configuration.
 type config struct {
-	minSize          int
-	level            int
-	writer           writer.GzipWriterFactory
-	contentTypes     func(ct string) bool
-	keepAcceptRanges bool
-	setContentType   bool
-	suffixETag       string
-	dropETag         bool
-	jitterBuffer     int
-	randomJitter     string
-	sha256Jitter     bool
+	minSize                 int
+	level                   int
+	writer                  writer.GzipWriterFactory
+	contentTypes            func(ct string) bool
+	keepAcceptRanges        bool
+	setContentType          bool
+	suffixETag              string
+	dropETag                bool
+	jitterBuffer            int
+	randomJitter            string
+	sha256Jitter            bool
+	allowCompressedRequests bool
 }
 
 func (c *config) validate() error {
@@ -584,6 +603,15 @@ type option func(c *config)
 func MinSize(size int) option {
 	return func(c *config) {
 		c.minSize = size
+	}
+}
+
+// AllowCompressedRequests will enable or disable RFC 7694 compressed requests.
+// By default this is Disabled.
+// See https://datatracker.ietf.org/doc/html/rfc7694
+func AllowCompressedRequests(b bool) option {
+	return func(c *config) {
+		c.allowCompressedRequests = b
 	}
 }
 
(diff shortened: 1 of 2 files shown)


gzhttp/compress.go
```diff
@@ -464,6 +464,12 @@ func NewWrapper(opts ...option) (func(http.Handler) http.HandlerFunc, error) {
 	return func(h http.Handler) http.HandlerFunc {
 		return func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Add(vary, acceptEncoding)
+			if contentGzip(r) {
+				if readerGzipBody, err := gzip.NewReader(r.Body); err == nil {
+					r.Body = io.NopCloser(readerGzipBody)
+				}
```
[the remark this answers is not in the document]
klauspost: [zstdReader](https://github.com/klauspost/compress/blob/0deb85a2cb5821965f1d7bf334f4036eb501f06f/gzhttp/transport.go#L200-L204) is also available for easy implementation. (optional)

gzhttp/compress.go
```diff
@@ -464,6 +464,12 @@ func NewWrapper(opts ...option) (func(http.Handler) http.HandlerFunc, error) {
 	return func(h http.Handler) http.HandlerFunc {
 		return func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Add(vary, acceptEncoding)
+			if contentGzip(r) {
+				if readerGzipBody, err := gzip.NewReader(r.Body); err == nil {
+					r.Body = io.NopCloser(readerGzipBody)
```
klauspost, line 469: The header field must be removed to avoid upstream confusion.

klauspost (changes requested):
Please add and implement these options:

```
// AllowCompressedRequests will enable or disable RFC 7694 compressed requests.
// By default this is Disabled.
// See https://datatracker.ietf.org/doc/html/rfc7694
func AllowCompressedRequests(b bool) option {
	return func(c *config) {
		//....
	}
}

// SendAcceptEncoding will send Accept-Encoding on all responses as per RFC 7694.
// By default, none will be sent.
// Calling this with no parameters will send the currently supported encodings back.
// This does not affect which encodings are accepted.
func SendAcceptEncoding(enc ...string) option {
	return func(c *config) {
		//....
	}
}
```

While it would be really nice to have this enabled by default, I think the "zip-bomb" risk is too high.

klauspost:
Just as an example [AWS S3](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax) "overloads" `Content-Encoding` on their uploads. in their terms it means the object should remain untouched and always return `Content-Encoding: whatever` when the object is downloaded.  This change would be breaking for that if enabled by default.

gzhttp/compress.go
```diff
@@ -752,6 +758,12 @@ func RandomJitter(n, buffer int, paranoid bool) option {
 	}
 }
 
+// contentGzip returns true if the given HTTP request indicates that it gzipped.
+func contentGzip(r *http.Request) bool {
+	// See more detail in `acceptsGzip`
+	return r.Method != http.MethodHead && parseEncodingGzip(r.Header.Get(contentEncoding)) > 0
```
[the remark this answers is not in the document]
mirecl: done

gzhttp/compress.go
mirecl: done

mirecl:
> Just as an example [AWS S3](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html#API_PutObject_RequestSyntax) "overloads" `Content-Encoding` on their uploads. in their terms it means the object should remain untouched and always return `Content-Encoding: whatever` when the object is downloaded. This change would be breaking for that if enabled by default.

@klauspost 
+ ``AllowCompressedRequests`` - should help solve this problem
+  ``zstdReader`` - i suggest limiting ourselves to gzip support for now

gzhttp/compress.go
```diff
@@ -464,6 +464,12 @@ func NewWrapper(opts ...option) (func(http.Handler) http.HandlerFunc, error) {
 	return func(h http.Handler) http.HandlerFunc {
 		return func(w http.ResponseWriter, r *http.Request) {
 			w.Header().Add(vary, acceptEncoding)
+			if c.allowCompressedRequests && contentGzip(r) {
+				r.Header.Del(contentEncoding)
+				r.Body = io.NopCloser(&gzipReader{body: r.Body})
```
klauspost, line 469: No need to wrap. `*gzipReader` is an `io.ReadCloser` that will close the body. Also it doesn't leak the decoder.

```suggestion
				r.Body = &gzipReader{body: r.Body}
```

gzhttp/compress.go
```diff
@@ -464,6 +464,12 @@ func NewWrapper(opts ...option) (func(http.Handler) http.HandlerFunc, error) {
...
+				defer r.Body.Close()
```
klauspost, line 470:
```suggestion
				defer r.Close()
```
With the change above and this the body is always closed and decoder is recycled.

klauspost (changes requested):

gzhttp/compress_test.go
```diff
@@ -89,6 +89,25 @@ func TestMustNewGzipHandler(t *testing.T) {
 	handler.ServeHTTP(res3, req3)
 
 	assertEqual(t, http.DetectContentType([]byte(testBody)), res3.Header().Get("Content-Type"))
+
+	// send compress request body
+	handlerCompressedRequests := newTestHandlerLevel(testBody, AllowCompressedRequests(true))
```
klauspost, line 94: Please add a test with default settings and check if it remains compressed.

gzhttp/compress.go
mirecl: @klauspost, ``defer r.Close`` - not found this function, maybe we should leave the old version ``defer r.Body.Close()``?

klauspost: Ah. Yeah. Correct.

mirecl:
@klauspost, all tasks done! Thank you very much for your patience and tips)

klauspost:
Cool. Giving it a final review! Thanks for the contribution!

klauspost (approved):

PR #1007, opened 2024-09-18
zstd: Improve memory usage on small streaming encodes

klauspost:
Very small streams will use EncodeAll internally when closing and no header has been written.

This will pull a new encoder from the async buffer.

Instead re-use the stream encoder.

Before:
```
BenchmarkMem/flush-32         	    1359	    837989 ns/op	 7376959 B/op	     109 allocs/op
BenchmarkMem/no-flush-32      	     129	   8884753 ns/op	112044489 B/op	     254 allocs/op
```

After:
```
BenchmarkMem/flush-32         	    1254	    922593 ns/op	 7376966 B/op	     109 allocs/op
BenchmarkMem/no-flush-32      	    1488	    841270 ns/op	 7374164 B/op	      29 allocs/op
```

Test is pretty much worst case, but shows the issue nicely.


klauspost:
@coderabbitai review

PR #1011, opened 2024-09-27
No content-type on no body response code

juliens:
This PR addresses an issue where the Content-Type was being automatically detected and set, even for responses that are not allowed to have a body (e.g., status codes 204 No Content, 304 Not Modified, etc.).

## Summary by CodeRabbit

- **New Features**
	- Introduced a function to determine if a response status code allows for a body, enhancing response handling based on HTTP standards.
  
- **Bug Fixes**
	- Updated response handling to ensure the `Content-Type` header is absent for `204 No Content` status responses.

- **Tests**
	- Added a test to verify that no `Content-Type` header is sent for `204 No Content` responses, ensuring compliance with HTTP specifications.

### DIFF
--- a/gzhttp/compress.go
+++ b/gzhttp/compress.go
@@ -349,12 +349,12 @@ func (w *GzipResponseWriter) Close() error {
 			ce = w.Header().Get(contentEncoding)
 			cr = w.Header().Get(contentRange)
 		)
-		if ct == "" && bodyAllowedForStatus(w.code) {
+		if ct == "" {
 			ct = http.DetectContentType(w.buf)
 
 			// Handles the intended case of setting a nil Content-Type (as for http/server or http/fs)
 			// Set the header only if the key does not exist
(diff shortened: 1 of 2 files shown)


gzhttp/compress.go
```diff
@@ -335,7 +349,7 @@ func (w *GzipResponseWriter) Close() error {
 			ce = w.Header().Get(contentEncoding)
 			cr = w.Header().Get(contentRange)
 		)
-		if ct == "" {
+		if ct == "" && bodyAllowedForStatus(w.code) {
```
klauspost, line 352: Let's add the check below at `// Set the header only if the key does not exist`

klauspost (changes requested):
Let's keep the internal part as-is, but skip setting the header on these types.

klauspost (approved):

klauspost:
Thanks for the contribution!

PR #1023, opened 2024-11-12
zstd: Read doc typo

dezza:
should say "read"

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated documentation for the `Read` method to accurately reflect that it returns the number of bytes read.
	- Enhanced error handling in the `Reset` method to manage decoder states and nil readers more effectively.
	- Improved control flow in the `nextBlock` method to ensure error states are preserved during operations.


klauspost:
I would say both are fine, and tbh I don't really think it is unclear as it is..

> // Read bytes from the decompressed stream into p.
> // Returns the number of bytes written (into p) and any error that occurred.

PR #1059, opened 2025-03-12
zstd: Fix incorrect buffer size in dictionary encodes

klauspost:
Fix incorrect dLongTableShardSize leading to inefficient zeroing of

Also make shards 128 bytes to reduce memory use somewhat.

### DIFF
--- a/zstd/enc_base.go
+++ b/zstd/enc_base.go
@@ -8,7 +8,7 @@ import (
 )
 
 const (
-	dictShardBits = 6
+	dictShardBits = 7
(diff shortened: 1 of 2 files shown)


klauspost:
Reported by @cfif1982 - fuzz testing before merge.

klauspost:
```
fuzz: elapsed: 18h39m23s, execs: 905973345 (31266/sec), new interesting: 21 (total: 7898)
```

Seems stable. Merging.

PR #1069, opened 2025-03-27
typo : s2/index.go

wooffie:
I don't know, but maybe we should output like this?

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error messaging to ensure accurate feedback when encountering data order issues.


klauspost (approved):
👍🏼

PR #1086, opened 2025-07-21
Avoiding extra allocation in zlib.reader.Reset

travelpolicy:
Without change in this PR line 182 ( z.digest.Reset() ) never works and we always have an extra allocation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of internal state when resetting, ensuring certain data is preserved as expected.


klauspost (approved):
Good spot!

PR #1144, opened 2026-04-21
docs: use unix line-endings for markdown files

thaJeztah:
We noticed some churn in our CI where files ended up with the wrong line endings; opening this PR for consideration (I realise it's a large amount of "lines" changed, but hopefully a one-time rip of the bandaid is OK).

## Summary by CodeRabbit

* **Chores**
  * Added a `.gitattributes` rule to standardize markdown file line endings.
  * Normalized line endings across multiple README files.


klauspost:
> We noticed some churn in our CI where files ended up with the wrong line endings

What do you mean?

thaJeztah:
Maybe @maxcleme  recalls the exact details (it's in one of our internal repositories, where I noticed an extra job was added to fix line endings).

thaJeztah:
I should add that in general, vendored files should always be considered "opaque", so not sure if that job should be there in the first place, but I think LF is more commonly used for Markdown, so thought it was worth opening a pull request for consideration 😅.

maxcleme:
We had vendoring issue in the past with dependabot PR when tied to windows-specific dependencies that forced us to use the following as a workaround:

```yml
      - name: Convert files in vendor folder to Unix format
        run: |
          find vendor -type f -print0 | xargs -0 dos2unix
```

After a quick search, looks like it is already reported but not fixed yet. https://github.com/dependabot/dependabot-core/issues/8693

thaJeztah:
Thanks!

PR #1148, opened 2026-05-07
snappy/s2: Limit length of varint in `decodedLen`

eustas:
To match C++ implementation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stricter validation of compressed-data headers to better detect and reject malformed/overlong encodings, reducing risk of silent corruption.

* **Tests**
  * Added test cases covering malformed length encodings to ensure decoding surfaces corruption errors consistently.

### DIFF
--- a/s2/s2_test.go
+++ b/s2/s2_test.go
@@ -227,6 +227,15 @@ func TestInvalidVarint(t *testing.T) {
 		// varint] (up to a maximum of 2^32 - 1)".
 		"valid varint (as uint64), but value overflows uint32",
 		"\x80\x80\x80\x80\x10",
+	}, {
+		// 8-byte overlong varint encoding value 4. The format spec limits
+		// the varint preamble to 5 bytes. binary.Uvarint accepts up to 10
+		// bytes so this decodes to a valid uint32 value (4), but the
+		// overlong encoding must be rejected.
+		"overlong varint (8 bytes encoding value 4), exceeds 5-byte max",
+		"\x84\x80\x80\x80\x80\x80\x80\x00" +
+			"\x00\x00\x00\x00\x00\x00\x00" +
+			"\x30",
 	}}
 
 	for _, tc := range testCases {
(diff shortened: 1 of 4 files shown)


s2/s2_test.go
```diff
@@ -227,6 +227,15 @@ func TestInvalidVarint(t *testing.T) {
 		// varint] (up to a maximum of 2^32 - 1)".
 		"valid varint (as uint64), but value overflows uint32",
 		"\x80\x80\x80\x80\x10",
+	}, {
+		// 8-byte overlong varint encoding value 4. The format spec limits
+		// the varint preamble to 5 bytes. binary.Uvarint accepts up to 10
+		// bytes so this decodes to a valid uint32 value (4), but the
+		// overlong encoding must be rejected.
```
klauspost, line 234: The [spec](https://github.com/google/snappy/blob/main/format_description.txt) doesn't state that. It states:

> The stream starts with the uncompressed length (up to a maximum of 2^32 - 1),
> stored as a little-endian varint. Varints consist of a series of bytes,
> where the lower 7 bits are data and the upper bit is set iff there are
> more bytes to be read. In other words, an uncompressed length of 64 would
> be stored as 0x40, and an uncompressed length of 2097150 (0x1FFFFE)
> would be stored as 0xFE 0xFF 0x7F.

klauspost:
The [go version](https://github.com/golang/snappy/blob/master/decode.go#L33) uses the same mechanism.

I don't really see any reason to change it. If people want to waste bytes IMO they are free to do so.

eustas:
I'll update go version soon (need to unarchive repo first).

The purpose is consistency: we want all decoders to produce the same output from the same input.

The "golden" behavior is the one from C++ implementation. There we decode only until "offset" is less than 32 (given 7-bit increment that is up to 5 bytes). In brotli we have even more strict check that disallows leading zero nibs, but I can imagine some hardware encoder might prefer to have fixed size preamble if the stream length is unknown upfront, so checking for leading zeros would be harmful for snappy.

Yes, that is not exactly what spec says; going to update it to clarify that as well.

klauspost (approved):
OK. I will revert at the first sign of problems.

thaJeztah:
> I'll update go version soon (need to unarchive repo first).

To save others arriving here from finding the related changes; PR (pending) is here;

- https://github.com/golang/snappy/pull/77

eustas:
Will poke the owner soon.

klauspost:
It has been released as part of https://github.com/klauspost/compress/releases/tag/v1.19.0

PR #1149, opened 2026-05-08
gzhttp: match qvalue parameter case-insensitively (RFC 7231)

z9z:
parseCoding only matched lowercase q=, but RFC 7231 says the weight parameter name is case-insensitive (Q= should also work).
Switched to strings.EqualFold on the 2-byte prefix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP compression negotiation to properly handle quality value parameters regardless of case, ensuring requests with uppercase parameter names are processed correctly.


klauspost (approved):
lgtm

PR #1160, opened 2026-06-22
zstd: arm64 decoder asm, generated via avo arm64 lowering printer

lizthegrey:
## arm64 assembly for the zstd decoder

On arm64 the zstd decode hot path (FSE table build + the sequence decoder) currently falls back to pure Go, because the decoder asm is avo-generated and avo had no arm64 backend. This adds arm64 asm for the whole decoder:

- `fse_decoder_arm64.s` — `buildDtable`
- `seqdec_arm64.s` — `sequenceDecs_decode`/`_56`, `executeSimple`/`_safe`, `decodeSync`/`_safe`
- `fse_decoder_arm64.go`, `seqdec_arm64.go` — the Go shims (context structs + dispatch), mirroring the amd64 ones; `*_generic.go` build tags now exclude arm64.

### These files are generated, not hand-written

The arm64 asm is produced from the **existing, unmodified** amd64 avo generators (`_generate/gen.go`, `gen_fse.go`) via a new arm64 lowering printer for avo (mmcloughlin/avo#486). One `go generate` emits both arches from the same IR:

```
//go:generate go run gen.go -out ../seqdec_amd64.s -arm64 ../seqdec_arm64.s -pkg=zstd
```

so CI's existing generate-and-`git diff` job keeps amd64 and arm64 consistent. There is no hand-maintained 80 KB of arm64 asm to drift.

The decoder is scalar, so this needs **no NEON/SVE** (the only vector op is a `VLD1`/`VST1` 16-byte copy) and builds on **go1.26**.

### Dependency note

The printer isn't in an avo release yet, so `_generate/go.mod` pins it via `replace => github.com/honeycombio/avo@<sha>` (the fork keeps the `github.com/mmcloughlin/avo` module path, so the replace resolves). Once mmcloughlin/avo#486 merges and releases, the replace is dropped for a normal `require`. Note the compress **library** has no avo dependency at all — avo is generate-time only; the checked-in `.s` + shims build standalone.

### Performance (Apple M4, go1.26, asm vs `-tags noasm`)

| workload | asm | pure-Go | |
|---|---:|---:|---|
| `DecodeAll` kppkn.gtb.zst (sequence-heavy) | ~469 MB/s | ~304 MB/s | **~1.5×** |
| `DecodeAll` geo.protodata.zst | ~1225 MB/s | ~1155 MB/s | ~+6% |
| `buildDtable` (litLength/offset/matchLength) | — | — | ~+16–18% |

The win should be larger on narrower Graviton cores.

### Correctness

Full `go test ./zstd/...` decode + fuzz suite passes on `darwin/arm64` (CI's `macos-latest` is arm64, so this exercises the asm path, not just cross-compiles). Differential tests compare asm output against the pure-Go reference: `Test_seqdec_decoder`/`Test_seqdec_execute` (vs `seqs.zip`) and `TestBuildDtableARM64MatchesReference`. The printer itself has a differential execution harness in the avo PR (each generated function run against a Go reference on both arches).

Supersedes #1159 (the hand-written fse bootstrap).

## Summary by CodeRabbit

* **New Features**
  * Added optimized ARM64 implementations for Zstd sequence decoding and FSE decoding-table generation.
* **Bug Fixes**
  * Improved build/platform selection so the correct decoder paths are used for ARM64 and non-assembly targets.
  * Strengthened validation and error mapping for malformed inputs and insufficient output space.
* **Tests**
  * Added ARM64 reference-matching tests and benchmarks for FSE table generation.
  * Expanded fuzz testing to optionally compare assembly vs non-assembly behavior.
* **Chores**
  * Updated CI to run fuzzing/builds on an additional ARM runner.


zstd/seqdec_arm64.go
```diff
@@ -0,0 +1,337 @@
...
+
+	s.seqSize = 0
+	startSize := len(s.out)
+
+	var errCode int
+	if useSafe {
+		errCode = sequenceDecs_decodeSync_safe_arm64(s, br, &ctx)
+	} else {
+		errCode = sequenceDecs_decodeSync_arm64(s, br, &ctx)
+	}
```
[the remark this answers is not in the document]
lizthegrey: Not applicable here — the callers guard `nSeqs == 0` before ever invoking these functions, so the asm is never entered with `n == 0`:

- `decodeSync`: `blockDec` returns at blockdec.go:487-490 (`if hist.decoders.nSeqs == 0 { … return nil }`) before the call at :492.
- `decode`: returns at blockdec.go:660-663 before the call at :667.

This matches the existing **amd64** path — `seqdec_amd64.go` computes `iteration` the same way and relies on the same caller guards; this file is a faithful mirror of it. Adding an `nSeqs > 0` check only in the arm64 shim would diverge it from the amd64 shim to protect a path the callers already make unreachable, so I have not applied it.

lizthegrey:
Benchmarked the sequence-decode hot path (`Benchmark_seqdec_decodeSync`) on real Graviton, comparing this PR's arm64 asm against the pure-Go path (`-tags=noasm`, i.e. what arm64 ships today). `go1.26.4 linux/arm64`, single core (`taskset -c 1`), `-count=10`, via `benchstat`. Package tests (`go test -short ./zstd/`) pass on both.

**Graviton2** (c6g.large, Neoverse-N1) — decodeSync geomean **−20.7%** sec/op; throughput 1.00 → 1.27 GiB/s (**+26%**)
**Graviton3** (c7g.large, Neoverse-V1) — decodeSync geomean **−28.1%** sec/op; throughput 1.55 → 2.16 GiB/s (**+39%**)

`0 B`/`0 allocs` unchanged on both. Representative blocks (sec/op, noasm → asm):

| block | Graviton2 | Graviton3 |
|---|---|---|
| n=14746 | 436.7µs → 323.2µs (−26%) | 246.9µs → 166.9µs (−32%) |
| n=8028  | 321.4µs → 224.8µs (−30%) | 214.6µs → 141.9µs (−34%) |
| n=2898  | 122.1µs → 82.4µs (−33%) | 75.5µs → 46.0µs (−39%) |
| n=1000  | 45.5µs → 32.7µs (−28%) | 25.4µs → 17.9µs (−30%) |

The only regressions are degenerate tiny-sequence blocks where asm setup overhead dominates: `n-2` (2 sequences) +73% on G2 / +13% on G3, and `n-90` (90 seqs) +14% on G2 / −4% on G3. Every block with a realistic sequence count improves 21–39%. Happy to share the full benchstat output.

Co-authored-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

lizthegrey:
This is running in production at Honeycomb with no issues.

klauspost:
Amazing. I am preparing a new bigger release, so I will prioritize getting this in before release!

zstd/fse_decoder_arm64.s
```diff
@@ -0,0 +1,130 @@
+// Code generated by command: go run gen_fse.go -out ../fse_decoder_amd64.s -arm64 ../fse_decoder_arm64.s -pkg=zstd. DO NOT EDIT.
```
klauspost: The output doesn't appear to be [asmfmt](https://github.com/klauspost/asmfmt) formatted.

Either 

A) Fix the upstream, so it produces asmfmt output or
B) Add asmfmt as a go tool to the mod file and run a secondary format.

zstd/seqdec_arm64.go
```diff
@@ -0,0 +1,337 @@
+//go:build arm64 && !appengine && !noasm && gc
```
klauspost, line 1: Since the arm64 has so strong ties to the amd64 assembly wouldn't it be possible to combine these to a larger degree than having duplicates?

zstd/seqdec_arm64.s
```diff
@@ -0,0 +1,2408 @@
+// Code generated by command: go run gen.go -out ../seqdec_amd64.s -arm64 ../seqdec_arm64.s -pkg=zstd. DO NOT EDIT.
```
klauspost: asmfmt again

klauspost (commented):
Overall fine. To exercise the fuzzers a bit more, let' them run for a few hours.

Modify `FuzzDecodeAll` to compare assembly versus non-assembly. You can use a const in the function that enables this similar to `FuzzEncoding` - or if you feel like it add a commandline flag that enables this mode.

Run `FuzzEncoding` for an extended time. You can apply this to get more executions on the decoder:

```diff
diff --git a/zstd/fuzz_test.go b/zstd/fuzz_test.go
index 78e0066..933252a 100644
--- a/zstd/fuzz_test.go
+++ b/zstd/fuzz_test.go
@@ -132,10 +132,10 @@ func FuzzEncoding(f *testing.F) {
 	const (
 		// Test a subset of encoders.
 		startFuzz = SpeedFastest
-		endFuzz   = SpeedBestCompression
+		endFuzz   = SpeedFastest
 
 		// Also tests with dictionaries...
-		testDicts = true
+		testDicts = false
 	)
 
 	var dec *Decoder
```

Please add `ubuntu-24.04-arm` to the CI runners.

Add a arm64 to the CI, similar to the existing: https://github.com/klauspost/compress/blob/master/.github/workflows/go.yml#L146-L175 - if it makes sense to add platform matric do that, but you are also welcome to copy it.

This should give a good confidence for this. I don't have arm64 HW myself, so I can't test directly.

zstd/fse_decoder_arm64.s
lizthegrey: Fixed upstream (option A): the avo arm64 printer now buffers each block and aligns operands to a common column, and emits a blank line before labels — matching avo's amd64 (goasm) layout. Regenerated here, and re-pinned the avo fork to that commit.

`asmfmt -d` now reports **zero non-comment differences** on both arm64 files — the output is formatted identically to the amd64 files. The only residual `asmfmt` flags is the standalone-`// comment` indentation, and it's identical on the amd64 files (goasm tab-indents these): e.g. `seqdec_amd64.s` shows 92 such lines, `seqdec_arm64.s` 48, `fse_decoder_*` 0 — all comment-only, no operand/layout differences. So arm64 is now exactly as asmfmt-clean as the existing amd64 output, without adding a secondary format step (which would also reformat the amd64 comments). If you'd prefer the comments at column 0 too, that's a goasm-level change affecting both arches.

zstd/seqdec_arm64.s
lizthegrey: Addressed — see above: operands are now column-aligned by the avo printer, matching the amd64 output; `asmfmt` shows only the same comment-indentation residual goasm produces on amd64.

zstd/seqdec_arm64.go
lizthegrey: done.

zstd/fuzz_test.go
```diff
@@ -50,6 +58,26 @@ func FuzzDecodeAll(f *testing.F) {
...
+			// output, exercising the assembly against the non-assembly path.
+			restore := cpuinfo.DisableBMI2()
+			decRef, errRef := NewReader(nil, WithDecoderLowmem(true), WithDecoderConcurrency(1), WithDecoderMaxMemory(20<<20), WithDecoderMaxWindow(1<<20), IgnoreChecksum(true))
+			if errRef != nil {
+				restore()
+				t.Fatal(errRef)
+			}
+			bRef, errRef := decRef.DecodeAll(b, make([]byte, 0, len(b)))
+			decRef.Close()
+			restore()
```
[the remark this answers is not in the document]
lizthegrey: Applied in ec89cf1 — switched to `defer cpuinfo.DisableBMI2()()` (and `defer decRef.Close()`), matching the existing `FuzzDecAllNoBMI2` idiom so BMI2 is re-enabled on every exit path.

lizthegrey:
## Fuzzing + CI follow-ups

**FuzzDecodeAll now compares assembly vs non-assembly** (const-gated): on amd64 it decodes again with BMI2 disabled and requires byte-identical output; on arm64 (`HasBMI2`=false) that in-process check is skipped, so the asm-vs-generic differential is covered by also running with `-tags=noasm` over the shared corpus.

**Local fuzzing (Apple M4, ~37 min total, all clean, 0 crashers):**
- `FuzzDecodeAll` (arm64 asm): 18.8M + 56M execs
- `FuzzDecodeAll` `-tags=noasm` (generic, over the asm-grown shared corpus): 14.2M
- `FuzzDecoder` (arm64 asm)
- `FuzzDecodeAll` amd64 (BMI2 vs non-BMI2 comparison): 7.6M
- `FuzzEncoding` with the fast encoder per your patch (round-trips exercise the decoder): 37.5M

**A multi-hour run is now going on real arm64 hardware** (Cortex-A72, 16 cores, ECC RAM) as a background service — `FuzzDecodeAll` then `FuzzDecoder`, 3h each. Will report if anything turns up.

**CI:** added `ubuntu-24.04-arm` to the build matrix (vet/test/noasm/race across Go versions on Linux arm64) and to the `fuzz-zstd` job, so the arm64 assembly is now exercised directly in CI rather than only cross-compiled.

klauspost:
Amazing! I will review tomorrow!

zstd/fse_decoder_arm64.s
lizthegrey: One follow-up worth noting (out of scope here): after the operand-alignment fix, `asmfmt -l` still flags `seqdec_arm64.s` — but **only** for standalone-comment indentation (`\t//` → `//`), with zero operand/layout differences. The existing `seqdec_amd64.s` is flagged identically (goasm tab-indents those comments on both arches). I left it matching amd64 rather than diverging arm64 alone.

If you want both arches strictly `asmfmt -l`-clean, that is a goasm-level comment-emission change (or an `asmfmt -w` generate step over both `.s` outputs) — happy to do it as a separate PR so it applies uniformly to amd64 and arm64 instead of just touching the new file.

klauspost (approved):
lgtm!

lizthegrey:
Multi-hour run finished clean on real arm64 (Cortex-A72, 16 cores, ECC):

- `FuzzDecodeAll` (arm64 asm), 3h — **PASS**, 454.7M execs, 0 crashers
- `FuzzDecoder` (arm64 asm), 3h — **PASS**, 0 crashers

No failures over the full 6 hours. Thanks for merging!

klauspost:
@lizthegrey Thanks! When I get a little time i will try it out on [minlz](https://github.com/minio/minlz), which has a lot of scalar avo assembly!

lizthegrey:
> @lizthegrey Thanks! When I get a little time i will try it out on [minlz](https://github.com/minio/minlz?rgh-link-date=2026-06-30T16%3A45%3A23Z), which has a lot of scalar avo assembly!

Now we just gotta get the patch accepted upstream in avo! :)

PR #1164, opened 2026-06-30
Merge commit from fork

klauspost:
`int(r)` would wrap negative for r > MaxInt64, slipping past the bounds check and causing an OOB read in encode.

Fix & add regression test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of dictionary repeat values to prevent overflow-related out-of-bounds reads during encoding.
  * Ensured extremely large repeat values are rejected safely, while valid boundary values continue to work as expected.

* **Tests**
  * Added coverage for repeat-value overflow cases and boundary handling.


klauspost:
(already merged, just running CI)

PR #1169, opened 2026-07-15
flate: use `Peek` instead of `ReadByte` for the `bufio.Reader` decode path

joechenrh:
## Problem

`huffmanBufioReader` fills the bit buffer with `fr.ReadByte()` per input byte. `bufio.Reader.ReadByte` doesn't inline, so each byte is a call barrier that spills the fill loop's registers — ~1.85× slower than the `*bytes.Reader` variant, whose `ReadByte` inlines to a slice index.

## Change

Read via `Peek` instead: index bufio's buffered slice directly (`pbuf[pos]`), refilling when drained. `Peek` only views the buffer, so consumed bytes are `Discard`ed before every return, keeping the reader positioned for `nextBlock`/`moreBits`/the gzip trailer. `peekBufio` refills from at most one `Read` (not `Peek(Size())`, which would block a sync writer).

Only the `*bufio.Reader` variant changes (in `_gen/gen_inflate.go`); the other four stay byte-identical.

## Validation

- flate/gzip/zip suites + `-race` pass on linux/amd64, linux/386, darwin/arm64.
- `FuzzInflateBufio` (forces the bufio path, byte-exact round-trip): ~110k execs, no mismatch.

## Numbers

`*bufio.Reader` decode, 8 MiB literal-heavy (median of 3):

| arch | before | after |
| --- | --- | --- |
| linux/amd64 | 128 MB/s | 151 MB/s (~1.18×) |
| darwin/arm64 | 219 MB/s | 287 MB/s (~1.31×) |

Match-heavy data ~1.05–1.07× (huffman-table-load bound, unchanged).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compressed-data decompression when input arrives in very small or irregular buffered reads.
  * Ensured decompressed output remains accurate across buffered-reader boundary conditions and error scenarios.

* **Tests**
  * Added fuzz testing covering compressed and decompressed data across multiple compression levels and reader behaviors.

### DIFF
--- a/flate/inflate_gen.go
+++ b/flate/inflate_gen.go
@@ -59,7 +59,6 @@ readLiteral:
 					if err != nil {
 						f.b, f.nb = fb, fnb
 						f.err = noEOF(err)
-
 						return
 					}
 					f.roffset++
@@ -120,7 +119,6 @@ readLiteral:
 						fmt.Println("morebits n>0:", err)
 					}
 					f.err = err
-
 					return
 				}
 				f.roffset++
@@ -149,7 +147,6 @@ readLiteral:
 						fmt.Println("morebits f.nb<5:", err)
 					}
 					f.err = err
-
 					return
 				}
 				f.roffset++
@@ -174,7 +171,6 @@ readLiteral:
 					if err != nil {
 						f.b, f.nb = fb, fnb
 						f.err = noEOF(err)
-
 						return
 					}
 					f.roffset++
@@ -219,7 +215,6 @@ readLiteral:
 						fmt.Println("morebits f.nb<nb:", err)
 					}
 					f.err = err
-
 					return
 				}
 				f.roffset++
@@ -315,7 +310,6 @@ readLiteral:
 					if err != nil {
 						f.b, f.nb = fb, fnb
 						f.err = noEOF(err)
-
 						return
 					}
 					f.roffset++
@@ -376,7 +370,6 @@ readLiteral:
 						fmt.Println("morebits n>0:", err)
 					}
 					f.err = err
-
 					return
 				}
 				f.roffset++
@@ -405,7 +398,6 @@ readLiteral:
 						fmt.Println("morebits f.nb<5:", err)
 					}
 					f.err = err
-
 					return
 				}
 				f.roffset++
@@ -430,7 +422,6 @@ readLiteral:
 					if err != nil {
 						f.b, f.nb = fb, fnb
 						f.err = noEOF(err)
-
 						return
 					}
 					f.roffset++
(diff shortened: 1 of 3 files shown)


flate/inflate_gen.go
```diff
@@ -50,6 +59,7 @@ readLiteral:
 					if err != nil {
 						f.b, f.nb = fb, fnb
 						f.err = noEOF(err)
+
```
klauspost: Can we get rid of these empty lines?

klauspost (commented):
Did you investigate other types? 

`*bytes.Reader` doesn't have it directly, but could be tricked into giving the backing bytes with [WriteTo](https://pkg.go.dev/bytes#Buffer.WriteTo) - which in it's current implementation will send all bytes as a single write. [Seek](https://pkg.go.dev/bytes#Reader.Seek) can be used to forward.

`*bytes.Buffer` is less obvious. You can of course easily get the underlying bytes, but forwarding it cheaply without doing a Reset (which could break callers) seems more tricky.

Of course if ReadByte is inlined it is probably not worth pursuing.

flate/inflate_gen.go
joechenrh: Fixed in https://github.com/klauspost/compress/commit/21318cfdf964e1e3c99f6ebea477aaf6809d8a7e — the `$GETBYTE_E$` placeholder had a leading newline the template already supplied, so every read site picked one up.

joechenrh:
Yes — I benchmarked each reader type with a flate-style fill loop (concrete type per func, mirroring the codegen), MB/s:

| reader | fill loop |
| --- | --- |
| `*bufio.Reader` (ReadByte call) | ~465 |
| interface `Reader` | ~490 |
| `*bytes.Reader` (ReadByte inlined) | ~895 |
| `*bufio.Reader` + Peek (this PR) | ~955 |

The bytes types are already ~2× faster than bufio — `-gcflags=-m` confirms `inlining call to bytes.(*Reader).ReadByte`, while `bufio.(*Reader).ReadByte` isn't inlinable at all. I haven't implemented a backing-slice variant for them, so I won't put a number on what `WriteTo`/`Seek` would actually buy.

`huffmanGenericReader` is the other slow one (~490), but I think the doc contract rules it out: *"If r does not also implement `io.ByteReader`, the decompressor may read more data than necessary"* — for a `ByteReader` we promise **not** to over-read, so buffering or peeking it would break that.

That's also why this PR stays on `*bufio.Reader`: peeking bytes it has already buffered reads nothing extra — `peekBufio`'s `Peek(1)` triggers exactly the fill `ReadByte` would have, and `Peek(Buffered())` reads none. Read-neutral, which is what the sync tests lean on.

klauspost (approved):

PR #645, opened 2022-07-17
zstd: Fix crash on amd64 (no BMI) + Go fuzz test

klauspost:
Port zstd fuzz tests to Go 1.18 fuzz tests.

(May need time extension on Go 1.18 CI)

PR #1171, opened 2026-07-18
zstd: Re-enable unsafe decodeSync memory copies (#1168)

### LINKED ISSUE #1168: zstd: decodeSync's unsafe path has been dead code since 2022 (#644) — remove or fix?
While working on the arm64 lowering for #1160/#1167 I noticed `(*sequenceDecs).decodeSyncSimple` in `zstd/seqdec_asm.go` has:

```go
// FIXME: Using unsafe memory copies leads to rare, random crashes
// with fuzz testing. It is therefore disabled for now.
const useSafe = true
```

with the real dynamic logic (computing `useSafe` from buffer slack, same shape as `executeSimple`'s still-active dynamic `useSafe`) commented out below it. This dates to #644 (2022-07-17), soon after #637 added the faster unsafe copy (up to ~1.46x on some benchmarks) and shortly after #562/#563 (an earlier decodeSync crash, root-caused and fixed). #644 doesn't link an issue or root cause — so this reads as "another crash turned up, we cut losses" rather than a one-off.

Four years later, `sequenceDecs_decodeSync_{amd64,bmi2}` are still generated and shipped, but `useSafe` being a compile-time `true` means nothing ever calls them — no test or fuzz target exercises them. The arm64 avo lowering (#1160) mechanically translates the same IR, so this dead, untested path now exists on a second architecture too (`_arm64`/`_bmi2`), doubling the footprint of code nobody runs.

Given the track record (fixed once, still had to be disabled again within months), I don't think restoring the commented-out logic as-is is right either. Two reasonable paths:

1. **Remove it** — drop the unsafe generation from `_generate/gen.go`, simplify the `safe bool` out of `decodeSyncAsm`'s dispatch, drop the dead code in `seqdec_asm.go`. Mechanical, no behavior change from what ships today.
2. **Root-cause and re-enable it** with real regression coverage, matching how `executeSimple`'s analogous path stayed dynamically toggled.

Happy to send a PR either way — let me know which you'd prefer.

lizthegrey:
Closes #1168 by taking option 2 (root-cause + re-enable with regression coverage).

## Background

`(*sequenceDecs).decodeSyncSimple` has had `const useSafe = true` hardcoded since #644 (2022-07-17), so the `sequenceDecs_decodeSync_{amd64,bmi2,arm64}` asm variants are generated and shipped but never called. #644's comment blames "rare, random crashes with fuzz testing" from the extended (16-byte-block) memory copies.

## There were two defects, not one

The extended copies overrun the end of a literal run or match by up to 15 bytes by design, relying on correct decoded offsets/lengths and 16 bytes (`compressedBlockOverAlloc`) of slack on the output/literal buffers. Two independent bugs could break that contract:

**1. The bitReader overread (fixed in 2022).** An unguarded `bitReader` overread in `updateLength` produced out-of-range match offsets/lengths, with only debug-only asserts to catch them. Fixed three days after the disable in #645 (`4b4f3c9`), which added runtime overread guards and introduced the Go fuzz corpus. This PR's first commit restored the dynamic `useSafe` selection on that basis.

**2. A missing margin in the space check (found and fixed in this PR).** The original analysis above missed a case, and this PR's own arm64 CI caught it: a SIGSEGV in `runtime.scanobject` — the GC tripping over corrupted heap. The per-sequence space check requires only `outPos+ll+ml <= cap(s.out)`, with no margin for the copies' 15-byte overrun. A well-formed decode ends at `cap-16` and the overrun lands in slack — but a **malformed frame whose declared content size is a few bytes below what its sequences actually produce** ends inside the slack: the check passes, and the final block copy writes past `cap(s.out)`.

This, not the overread, is the likely source of #644's "rare, random" crashes: the ≤15-byte overrun lands in size-class slop that neither `-race` nor `-asan` poisons, so it only intermittently corrupts an adjacent live object and surfaces much later as a GC crash rather than at the faulting write.

The fix reserves `compressedBlockOverAlloc` in the space check on the unsafe path only (one `ADDQ` per unsafe variant; the bounds-exact safe variants keep the tight check). Well-formed streams are unaffected; an over-producing stream now returns `error_not_enough_space` instead of corrupting memory.

## Regression coverage

- **`TestDecodeSyncUnsafeOOB`** — the guard that actually catches this class. It drives the unsafe asm with real captured sequences into a canary-guarded buffer carved from a larger backing array, sized to model the under-reporting case, and asserts (via the extracted `useSafeDecodeSync()` helper) that the unsafe variant is really selected. It fails on the unpatched assembly (canary overwritten past `cap`) and passes with the fix, on amd64 and arm64.
- **`fuzz-zstd-asan` CI job** (CGO+clang, amd64+arm64) over `FuzzDecodeAll`/`FuzzDecAllNoBMI2`. Caveat learned the hard way: asan does **not** see sub-size-class slice overruns (they land in unpoisoned slop), so it guards gross overruns and allocation regressions, not this ≤15-byte class — that's what the canary test is for.

## Validation (amd64 + arm64/Cortex-A72, go1.25 + go1.26)

- Full `zstd` test suite, all build-tag combos, both architectures.
- asm-vs-noasm differential fuzz: byte-identical output.
- Canary reproducer: fail-before/pass-after verified on both architectures.
- ASan over the full decode corpus + 2.4M mutation execs: zero reports; plain fuzz 22.8M+ execs clean (default and `-tags=nounsafe`).

## Performance

`BenchmarkDecoder_DecodeAll` on arm64 (Cortex-A72), n=6, p=0.002:

- **+8.7% geomean throughput**, up to **+16%** on text (asyoulik +16%, plrabn12 +15%, alice29 +15%, kppkn +12%).
- No change on already-fast inputs (jpeg), as expected.
- The added space-check margin is one register add per sequence batch; no measurable cost.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved zstd `decodeSync` safety by dynamically selecting between safer and optimized copy behavior based on output/literals geometry.
  - Tightened output-capacity checks in the decode assembly paths to reduce the risk of out-of-bounds writes.
- **Tests**
  - Added an amd64/arm64 regression test that reproduces an unsafe/extended-copy scenario and verifies no memory beyond the output buffer is modified.
- **Chores**
  - Added an AddressSanitizer-protected CI fuzzing job for zstd decode fuzz targets.


lizthegrey:
the irony: "enable unsafe, it's safe now!" crashes on fuzz failure on the very first CI build.

lizthegrey:
Pushed a fix for the arm64 CI crash (`ba6f83c`).

**Root cause** — not the bitReader overread, but a missing copy margin. The per-sequence space check in `executeSingleTriple` requires only `outPos+ll+ml <= cap(s.out)`, but the unsafe extended copies write in 16-byte blocks and overrun the logical end by up to 15 bytes. When a decode ends inside the 16-byte over-allocation slack — e.g. a malformed frame whose declared content size is a few bytes under what its sequences produce — the check passes yet the final block copy writes past `cap(s.out)`. Go's size-class rounding hides the ≤15-byte overrun from `-race` and even `-asan` (it lands in unpoisoned slop), so it only intermittently corrupts an adjacent live object and surfaces later as a GC crash. That matches the "rare, random crashes" #644 originally disabled this path for.

**Fix** — reserve `compressedBlockOverAlloc` in the space check on the unsafe path only. A well-formed decode ends at `cap-16` and is unaffected; an over-producing stream now returns `error_not_enough_space` instead of corrupting memory. The safe copies stay bounds-exact with the tight check. One `ADDQ` per unsafe variant; safe variants unchanged.

**Repro/regression** — `TestDecodeSyncUnsafeOOB` drives the unsafe asm with the real captured sequences into a canary-guarded buffer sized to model the under-reported case. It fails (canary overwritten past `cap`) on the unpatched asm and passes with the fix, on both amd64 and arm64. Also validated with the full suite + `FuzzDecodeAll` (default and `-tags=nounsafe`) on go1.25 and go1.26.

Since asan can't see this overrun class, the `fuzz-zstd-asan` job gives false confidence here — the canary test is the real guard and belongs in CI.

zstd/seqdec_oob_test.go
```diff
@@ -0,0 +1,117 @@
...
+			for i := capOut; i < len(backing); i++ {
+				if backing[i] != canaryByte {
+					t.Fatalf("%s k=%d: unsafe decodeSync wrote past cap(out)=%d (first corrupted byte at +%d)",
+						tt.Name, k, capOut, i-capOut)
+				}
+			}
+		}
+	}
+	if asmPaths == 0 {
+		t.Fatal("test did not exercise the unsafe decodeSync asm path")
```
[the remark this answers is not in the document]
lizthegrey: Finding verification: the test's buffer geometry (cap = maxSyncLen+16, literals allocated with slack in readDecoders) did select the unsafe variant — but only implicitly. Nothing would catch a future change to the useSafe conditions or the harness allocation that flipped it to the safe copies; the test would keep passing while guarding nothing.

Fix (minimal, drift-proof): rather than duplicating the selection conditions in the test (which could itself drift), I extracted decodeSyncSimple's three-condition selection into useSafeDecodeSync() — a pure refactor, the explanatory comment moved with it, no behavior change — and the test now calls that same helper per buffer configuration, failing loudly with "buffer sizing selects the safe copy variant" if the unsafe path wouldn't be exercised. Canary validation and the asmPaths assertion are untouched, addressing the review's "preserve existing validation" requirement.

zstd/_generate/gen.go
```diff
@@ -1119,6 +1124,14 @@ func (e executeSimple) executeSingleTriple(c *executeSingleTripleContext, handle
...
 			LEAQ(Mem{Base: ll, Index: ml, Scale: 1}, baseAfterCopy)
 			ADDQ(c.outBase, baseAfterCopy)
+			if !e.safeMem {
+				// The extended copies below write in 16-byte blocks and overrun
+				// the logical end by up to compressedBlockOverAlloc-1 bytes.
+				// Reserve that margin so the overrun stays within cap(s.out);
+				// otherwise a stream whose decoded length lands inside the
+				// over-allocation slack passes this check yet writes past cap.
+				ADDQ(U8(compressedBlockOverAlloc), baseAfterCopy)
+			}
```
klauspost, line 1134: We should be able to do it for "free" in the LEAQ above:

```go
			addMargin := 0
			if !e.safeMem {
				addMargin = compressedBlockOverAlloc
			}
			LEAQ(Mem{Base: ll, Index: ml, Scale: 1, Disp: addMargin}, baseAfterCopy)
```

lizthegrey: Done.

klauspost (approved):

lizthegrey:
Done in 47ea5b1 — folded `compressedBlockOverAlloc` into the LEAQ displacement as you suggested. Only the unsafe amd64/bmi2 variants change (one fewer instruction each); the safe path (`addMargin=0`) and arm64 codegen are unaffected. Verified via regenerated asm diff + full `zstd` suite (incl. `TestDecodeSyncUnsafeOOB`) on both amd64 and arm64.

klauspost (approved):
lgtm

PR #1172, opened 2026-07-18
huff0: add arm64 assembly for Decompress4X/1X via avo lowering

lizthegrey:
This adds arm64 assembly for huff0's `Decompress4X`/`Decompress1X`, generated from the existing avo source with the arm64 lowering printer (`-arch amd64,arm64`) that already produces zstd's seqdec/fse arm64 asm (#1160/#1167). **The amd64 assembly output is byte-for-byte unchanged** (only the generated-by header and the build-constraint line move).

### Structure

* `huff0/_generate/gen.go` now emits both architectures from the one source; the `amd64` term moves out of `ConstraintExpr` (the `_amd64.s`/`_arm64.s` filename suffixes select the arch).
* The Go wrappers move to `decompress_asm.go`, shared by amd64 and arm64. Per-arch files keep only the asm stubs and dispatch: amd64 retains its BMI2 selection for `Decompress1X`; arm64 lowers the generic twin.
* `decompress_generic.go` is now the `(!amd64 && !arm64) || …` fallback.

### avo pin

The pinned honeycombio/avo lowering grows two commits (honeycombio/avo `lizf.arm64-lowering-printer`):

1. **Bug fix — exact x86 partial-register semantics for MOVB/MOVW.** The previous lowering wrote full registers where x86 writes one byte: `MOVB src, AH` clobbered the whole register, byte/word loads sign-extended. zstd's generators never tripped this (their byte moves land in registers with dead upper bits), but huff0 packs output bytes through AH and reads 16-bit table entries — every stream decoded wrong. Byte/word register destinations now lower to bit-field inserts (`BFI`)/zero-extending loads.
2. New lowerings huff0 needs (`XORL`, `SETcc`, `ADCB $0` carry-accumulate via `CSINC`, wrap-exact `ADDB`, `BSWAPL`), plus constant-folding of BMI2 control registers (`MOVQ $const` + `BEXTRQ` → single `UBFX`).

Both commits carry differential execution tests (`tests/arm64lower`) run against native semantics on amd64 and the lowered output on arm64.

`zstd/seqdec_arm64.s` is regenerated with the fixed printer (its `MOVB` sites now use the exact insert forms); its amd64 twin is unchanged.

### Benchmarks

16-core Cortex-A72 (ARMv8.0, Neoverse numbers TBD), go1.26.5, `-count 2`, serial runs. huff0 asm vs the previous pure-Go fallback:

| benchmark | Go | asm | Δ |
|---|---|---|---|
| Decompress1X/twain/262143 | 114.0 MB/s | 153.2 MB/s | **+34%** |
| Decompress1X/digits/262143 | 128.6 MB/s | 161.2 MB/s | **+25%** |
| Decompress1X/low-ent/262143 | 139.6 MB/s | 173.8 MB/s | **+24%** |
| Decompress4X/twain/262143 | 183.5 MB/s | 205.2 MB/s | **+12%** |
| Decompress4X/digits/262143 | 222.3 MB/s | 222.4 MB/s | ±0% |
| Decompress4X/low-ent/262143 | 262.3 MB/s | 252.1 MB/s | −4% |

4X is content-dependent: the x86 AL/AH/BSWAP byte-packing lowers to BFI chains, so low-entropy (short-code) streams lose a little while higher-entropy streams gain. Net effect on zstd `BenchmarkDecoder_DecodeAll` (literals are 27–44% of these payloads):

| file | before | after | Δ |
|---|---|---|---|
| comp-data.bin | 151.7 MB/s | 162.1 MB/s | **+6.9%** |
| urls.10K | 249.8 MB/s | 257.4 MB/s | **+3.0%** |
| geo.protodata | 448.5 MB/s | 461.7 MB/s | **+2.9%** |
| html | 375.6 MB/s | 382.7 MB/s | **+1.9%** |
| kppkn.gtb | 198.0 MB/s | 200.8 MB/s | +1.4% |
| plrabn12.txt | 156.5 MB/s | 157.7 MB/s | +0.8% |

(Baseline includes #1168's dynamic-useSafe re-enable; fireworks.jpeg/paper-100k are literal-free-ish and unchanged.)

### Validation

* Full huff0 + zstd test suites on arm64 (Cortex-A72) and amd64, including the zstd decode fuzz seeds.
* avo differential execution tests for every new/changed lowering on both architectures.
* amd64 generated asm verified byte-identical before/after.

## Summary by CodeRabbit

* **New Features**
  * Added ARM64 support for Huffman decompression, enabling optimized decoding on ARM-based systems.
  * Added architecture-specific acceleration for supported decompression workloads.

* **Bug Fixes**
  * Improved ARM64 sequence decoding and small-copy handling for more reliable compressed-data processing.
  * Preserved fallback decoding support for platforms and build configurations without assembly acceleration.

* **Performance**
  * Improved decompression efficiency on supported amd64 and arm64 systems.


klauspost (approved):
lgtm

PR #1173, opened 2026-07-18
zstd, huff0: promote arm64 stack spills to registers

lizthegrey:
**Stacked on #1172** (huff0 arm64) — review/merge that first; this branch contains its commit plus one more.

The arm64 lowering printer runs after avo's x86 register allocation, so it inherits x86's 14-GP-register budget and faithfully reproduces every spill the allocator emitted — as RSP-relative loads and stores — even though arm64 has 30 GP registers sitting idle. In the zstd sequence-decode inner loop that is ~10 stack accesses per sequence.

This adds an **opt-in** avo pass (`-arm64-promote-stack-slots`) that keeps eligible frame slots in otherwise-unused registers (R19–R26) for the lifetime of the function, turning those hot-path loads/stores into register moves. A slot qualifies only when **every** reference to it is a whole-8-byte `MOVQ` to or from a register (or an immediate store); any partial-width access, address-of, indexed access, or read-modify-write use disqualifies it, so a promoted register is always a faithful stand-in for the slot's full 64 bits. The hottest slots win the eight registers; the rest stay on the stack.

R19–R26 are safe to clobber without save/restore: these are call-free `NOSPLIT` leaves, and Go's ABI treats every integer register except `g` as caller-clobbered across the call into them — the same assumption the existing lowering already makes for R0–R16.

### Effect

`zstd/seqdec_arm64.s` drops from 78 to 40 RSP references (the sequence-decode loop is the big win). `huff0` was already register-resident and is unchanged. **The amd64 assembly is byte-identical.**

Measured on a 16-core Cortex-A72 (go1.26, `-count 6`, all p<0.01):

| benchmark | Δ |
|---|---|
| Benchmark_seqdec_decodeSync (geomean) | **−1.46%** (up to −6.9%) |
| Decoder_DecodeAll/kppkn.gtb | **−2.18%** |
| Decoder_DecodeAll/alice29.txt | **−1.93%** |
| Decoder_DecodeAll/lcet10.txt | **−1.63%** |
| Decoder_DecodeAll/plrabn12.txt | **−1.41%** |
| Decoder_DecodeAll/geo.protodata | −0.82% |
| Decoder_DecodeAll/urls.10K | −0.74% |

Non-sequence-bound inputs (fireworks.jpeg, comp-data.bin) are unchanged, as expected — the win comes entirely from the sequence-decode loop.

### avo pin

The pinned honeycombio/avo lowering branch gains one commit (`printer/arm64: promote x86 stack spills to registers`). The pass is off by default; only these generators opt in via the flag in their `go:generate` lines.

### Validation

* Full huff0 + zstd test suites on arm64 (Cortex-A72) and amd64.
* ~3.8M-exec fuzzing of the promoted build: FuzzDecodeAll, FuzzDecoder, and FuzzDecodeAll under `-tags=nounsafe` — no failures.
* avo differential test `StackAccum` carries an accumulator and loop bound through promoted slots, checked against a pure-Go reference on both architectures.
* amd64 generated asm verified byte-identical.

Neoverse validation is still pending; the numbers above are Cortex-A72 only.

## Summary by CodeRabbit

* **New Features**
  * Added optimized ARM64 support for Huffman decompression.
  * Extended assembly-accelerated decompression across supported AMD64 and ARM64 builds.
  * Improved ARM64 handling for Zstandard sequence decoding and finite-state entropy operations.

* **Bug Fixes**
  * Strengthened decompression bounds checking and error handling.
  * Added safeguards to prevent writes beyond the declared output buffer.

* **Tests**
  * Added randomized regression coverage validating round-trip decompression and buffer safety.


klauspost:
Out of curiosity - which order are they taken in?

lizthegrey:
Re "which order are they taken in?" — traced it in `printer/arm64.go` (`stackSlotPromotions`, avo): it's a single forward pass counting clean whole-8-byte `MOVQ` references per frame slot (any partial-width/indexed/RMW access anywhere disqualifies that slot permanently). Eligible slots are sorted **descending by reference count, ties broken by ascending stack offset**, and the top 8 get `R19`–`R26` in that order. So "hottest" is a literal static reference count, not textual/declaration order, and the result is fully deterministic (the sort key has no remaining ties once offset breaks them).

Your question also exposed a real gap: the existing differential test (`StackAccum`) only ever has 2 competing slots, so it never exercised the >8-slot ranking/eviction path at all — it could have silently degraded to "promote nothing" without failing. Added `ManySlotAccum` (12 slots, strictly-decreasing reference counts, unambiguous ranking) to the avo branch; confirmed the generated arm64 asm now actually uses all of R19–R26 *and* still has stack references for the 4 evicted slots.

Since this came up: I benchmarked the pass on hardware other than the Cortex-A72 in the PR description — Graviton2 (Neoverse N1) and Graviton4 (Neoverse V2), both clean/controlled server instances, 10 runs each way:

| Core | `Benchmark_seqdec_decodeSync` geomean |
|---|---|
| Cortex-A72 | **-1.46% faster** (up to -6.9%) |
| Neoverse N1 (Graviton2) | +0.41% slower |
| Neoverse V2 (Graviton4) | +0.20% slower |

The win is real but doesn't generalize past Cortex-A72 — both independently-tested Neoverse generations (one older/narrower than A72, one newer/wider) show a wash-to-slight-regression instead, with no consistent pattern by core age or OOO width. Given that only one of three tested microarchitectures actually benefits, and this adds a whole opt-in avo pass plus a new CLI flag to maintain, I'm closing this rather than shipping a knob whose payoff turned out to be specific to one core and doesn't hold up elsewhere.

I've split the underlying avo change into its own draft PR (mmcloughlin/avo, stacked separately from the base arm64-lowering-printer RFC) with this data in its description, in case it's useful for future work — but not carrying it here.

PR #1175, opened 2026-07-25
zstd, huff0: improved arm64 lowering, and narrow two table loads to MOVWQZX

lizthegrey:
> **Rebased onto master now that #1176 has merged.** The 148 lines of frame-offset churn are gone from this diff, so what remains is only the lowering improvement, the `MOVWQZX` narrowing, and one build-tag line per file. Description updated to match; see [Relationship to #1176](#relationship-to-1176-merged) for the cross-check that fell out of the rebase.

Three changes, separable and individually justified.

## 1. Regenerated arm64 assembly (the original PR)

Same generators, re-run against an avo lowering that emits better arm64. Two addressing improvements:

* **128-bit moves use `FMOVQ`**, which takes a base+displacement operand directly. Previously `VLD1`/`VST1` needed the address materialized into a scratch register first, so `MOVOU disp(base), X` cost two instructions instead of one.
* **Scalar accesses fold `base+index` into the instruction** where the architecture allows it — the index must be unscaled or scaled by exactly the access width, with no displacement alongside.

Measured on Neoverse N1 (Ampere, go1.26, `-count 6`, all p<0.03):

| benchmark | Δ |
|---|---|
| `Decoder_DecodeAll` geomean | **−1.53%** |
| html.zst | −2.29% |
| kppkn.gtb.zst | −2.08% |
| huff0 `Decompress4X/twain` | −2.81% |

Cortex-A72 agrees: `Decoder_DecodeAll` geomean **−1.28%** (p<0.01). Two microarchitectures, same direction — unlike #1173, where an A72 win did not survive on Graviton.

**That measurement still applies**, and the rebase made the proof simpler than it was before. The arm64 assembly now differs from the benchmarked revision by exactly one line per file — the build tag from change 3 — and by zero instruction lines:

```
$ for f in zstd/seqdec_arm64.s huff0/decompress_arm64.s; do
    diff <(git show 405fdc9:$f) <(git show HEAD:$f) | grep '^[<>]' | grep -vc 'go:build'
  done
0
0
```

## 2. Two table loads narrowed to `MOVWQZX` (generator change, amd64 win)

This is why `_generate/gen.go` and the amd64 assembly are touched at all. With the frame churn rebased away, it is now the entirety of the amd64 diff.

Making the lowering's 16-bit load exact — x86 `MOVW` preserves the upper 48 bits, arm64's load zero-extends — costs an instruction, and that instruction reads the register it writes, turning an independent load in huff0's inner decode loop into a loop-carried dependency. Measured **+11.74%** geomean on `Decompress4X` on N1.

Nothing here was miscompiled by the old behaviour: at all forty affected sites the destination's upper bits are dead, so zero-extending was harmless *as applied*. The generators simply asked for more than they needed. `MOVWQZX` asks for what they actually use, and lowers back to a single folded load.

It is also the better x86 encoding independently: `MOVZX` avoids the `66h` operand-size prefix and the partial-register merge dependency — the same hazard that cost 11.74% on arm64. `MOVW mem, CX` writes only CX's low 16 bits, so the CPU merges with the prior `RCX`, creating a false dependency on a value the loop immediately consumes.

On amd64 huff0 `Decompress4XNoTable` (Ryzen 5 5600X, `-count 12`), the effect splits cleanly by input size:

| input size | Δ sec/op |
|---|---|
| 100 B (9 cases) | **no significant change** (p = 0.06–0.94) |
| 10 KB and 256 KB (19 cases) | **−6.6% to −14.1%** (all p = 0.000) |

The two binaries differed in exactly 32 lines, all `MOVW`→`MOVWQZX`.

Quoting the geomean over the whole set would be **−7.59%**, but that mixes the nine zero-effect cases in with the rest, so it understates the loop improvement and overstates the benefit on tiny inputs — the split above is the honest summary. The 100-byte cases being flat is also what rules out a drift artifact: a thermal or ordering effect would have moved them too.

## 3. Build constraints bracketed

One line per generated file: `//go:build arm64 && !noasm && gc` becomes `//go:build arm64 && (!noasm && gc)`.

A no-op for every constraint in this repository, all of which are pure `AND` chains. Without it, a constraint containing `||` would splice to `arm64 && !noasm || purego`, and since `&&` binds tighter that parses as `(arm64 && !noasm) || purego` — pulling arm64 assembly into non-arm64 builds.

## Relationship to #1176 (merged)

#1176 fixed the frame offsets by hand-verified reasoning about what the lowering *should* emit. This branch's pinned avo contains the same fix in the lowering itself, applied from round seven onward.

Those were arrived at independently, so the rebase is a real check rather than a formality: **regenerating from this branch's pin, on top of merged master, reproduces `seqdec_arm64.s` byte for byte.** The two did not merely produce compatible answers, they produced the same one. Every local in all four affected functions sits at ≥ 8(RSP), leaving the saved link register slot untouched:

```
$ awk '/^TEXT/{f=$2} match($0,/-?[0-9]+\(RSP\)/){o=substr($0,RSTART,RLENGTH-5)+0;
    if(!(f in m)||o<m[f])m[f]=o} END{for(k in m)print k,m[k]}' zstd/seqdec_arm64.s | sort
·sequenceDecs_decode_56_arm64(SB), 8
·sequenceDecs_decode_arm64(SB), 8
·sequenceDecs_decodeSync_arm64(SB), 8
·sequenceDecs_decodeSync_safe_arm64(SB), 8
```

The one commit in this branch that originally carried the `+8` (round seven) is now assembly-neutral, like the other pin bumps.

## avo pin

`zstd/_generate` and `huff0/_generate` move to honeycombio/avo `a995462` (branch `lizf.arm64-lowering-printer`).

That range is mostly hardening rather than anything visible here: fourteen rounds of adversarial review of the lowering found and fixed 43 ways it could emit plausible-looking but wrong assembly. **Only one of them was ever reachable from these generators** — the frame-offset bug, which is why it went out separately as #1176. For all the rest, regenerating and getting byte-identical output at each step is the standing check that a lowering change cannot affect shipped code.

Two structural changes worth flagging, since they reduce what the lowering will accept:

* **Preprocessor directives are refused.** The lowering used to evaluate `GOAMD64` conditionals inline; that machinery produced 11 of the 43 bugs, all of them cases where its model of when a directive is live disagreed with the assembler or the postprocessing step. No generator here emits directives, and BMI2 variants continue to work through twin *functions*, which is the mechanism these generators actually use.
* **Flag effects are checked, not assumed.** Every mnemonic the lowering can emit is classified as flag-writing or not; emitting an unclassified one, emitting a flag-writer from a lowering claimed neutral, or *failing* to emit one from an instruction marked as a flag producer are all now generation-time failures.

The differential test harness runs the same avo programs on both architectures against pure-Go references, with coverage gates that fail if any opcode or condition the lowering handles is never executed.

## Validation

Re-run after the rebase, not carried over:

* Full `zstd` and `huff0` suites pass on Neoverse N1 (go1.26.5, arm64) and on amd64, uncached.
* `go generate` reproduces every checked-in file exactly, from a clean tree — this is what establishes that the rebase resolution is the generator's own output and not a hand-merge.
* amd64 and arm64 both build and test with the bracketed constraints.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved compression and decompression efficiency on ARM64 and AMD64 platforms.
  * Optimized Huffman and sequence decoding paths for faster table lookups and memory operations.
  * Streamlined data copying in common decoding workflows.

* **Reliability**
  * Improved handling of decoded values during low-level operations, helping maintain consistent results across supported architectures.


lizthegrey:
Pushed `bd8f892`, bumping the pinned avo again. **The generated assembly is unchanged** — regenerating against the new pin reproduces every `.s` file byte for byte, so the PR's diff and its measurements are exactly as before.

The reason for the bump is a review pass over the lowering that turned up eight ways it could emit plausible-looking but wrong assembly. None is reachable from these generators — the byte-identical output is the proof — but the pin determines what the *next* regeneration uses, so it seemed worth carrying here rather than leaving for later.

Briefly, they fall into two groups. Wrong values: byte-extending moves read bits 7:0 from a high-byte source (AH renames to the same arm64 register as AL); a 32-bit conditional move emitted a 64-bit select, dropping x86's unconditional zero-extension of the destination; BEXTR with a memory source staged a control field in the same scratch register the address computation then overwrote; and compare/test/32-bit-multiply loaded eight bytes for a four-byte memory operand, which reads past the operand and can fault at a page boundary.

Wrong flags, all of which now fail generation rather than translate: a sub-word compare whose consumer sat past a flag-transparent MOV was wrongly treated as EQ/NE-only, so a signed byte compare lowered to an unsigned one; carry conditions after logical ops were inverted (x86 and arm64 both force carry to zero there, so the meaning-inverting map that is right after a compare is wrong after a TEST); INC/DEC preserve CF on x86 with no arm64 equivalent; and ADC's lowering assumed a borrow-producing predecessor, which is wrong after an addition.

lizthegrey:
Pushed again (`dfb95f0`) — pin now at avo `2b46a91`. **Generated assembly is still byte-identical**; the diff and the measurements in the description are unchanged.

A second, independent review pass over the already-fixed lowering found three more cases, which is why the pin moved once more. It first re-derived the previous round's findings from the pre-fix code and confirmed them fixed, then found these remaining:

- A carry-condition branch after an **addition** read inverted carry. x86 `ADD` and arm64 `ADDS` both set carry-out with the same sense, but the condition map translates by post-*compare* meaning, where the two architectures use opposite borrow conventions — so `JCS` after an `ADD` became `BLO`, taken exactly when the add did *not* carry. The previous round had guarded the logical, INC/DEC and ADC cases but not plain addition. Carry conditions now require a compare or subtract producer.
- The **XOR-self zeroing shortcut** returned a bare move, dropping the zero flag x86 sets there, so a following branch read stale flags.
- A **negative 32-bit immediate** sign-extended: `MOVL $-1, r32` leaves `0x00000000ffffffff` on x86, where the lowering set all 64 bits.

As before, none is reachable from these generators — the byte-identical regeneration is the evidence — and each is now covered by a test.

lizthegrey:
One more pin bump (`4604a59` → avo `6d9b258`); **generated assembly is still byte-identical**, so the diff and measurements stand.

This round came from having the reviewer verify the previous fixes rather than just accept them, which was worthwhile: one of them had over-corrected. Extracting a high-byte register faithfully on arm64 is only right where x86 can encode the same thing, and that is narrower than it looks — any REX-carrying encoding renames that operand to SPL, so `MOVBQZX AH, BX` actually assembles to `MOVZX SP, BX` and reads a stack byte. The previous fix made amd64 and arm64 compute *different* values from one program, which is precisely the divergence this lowering exists to prevent. Those forms now fail generation instead.

The other change is structural: the GOAMD64 conditionals are resolved in a pass before the flag analyses run, so those see exactly the instructions that will be emitted. Left as-is, a flag producer inside a dead arm could absorb the flag-setting mark a live producer before the conditional needed — silent stale flags, and plausible for s2, whose generated output carries dozens of such arms. Dead arms now lose their labels along with their code, so a jump into one is an assembler error rather than a silent landing after the conditional.

That is thirteen fixes across three review passes, none of them reachable from these generators — the byte-identical regeneration each time is the evidence — and each covered by a test.

lizthegrey:
Rewrote the description — this PR outgrew it, and two claims in the original are no longer true.

**Corrections to what I originally wrote here:** it said *"the amd64 assembly is byte-identical"* and *"regenerated arm64 assembly only — no source or algorithm change."* Both were accurate when opened and neither is now. `_generate/gen.go` changed in both packages, and the amd64 assembly changed with it. Flagging that explicitly rather than quietly editing, since anyone who read the original would have been misled about what to review.

What it contains now, separably:

1. **The original regen** — `FMOVQ` for 128-bit moves, folded `base+index` addressing. −1.53% `DecodeAll` geomean on N1, −1.28% on A72.
2. **Two table loads narrowed to `MOVWQZX`** — a generator change, and the reason amd64 moved. Worth **−7.59%** geomean on amd64 huff0 `Decompress4X` on its own merits (avoids the `66h` prefix and the partial-register merge dependency).
3. **The frame fix, which is #1176** — 148 lines of `(RSP)` offsets. Please take #1176 first; I will rebase and those lines leave this diff.
4. **Build constraints bracketed** — one line per file, a no-op for every constraint here, but correct if one ever contains `||`.

On the perf claim surviving all that churn: `seqdec_arm64.s` differs from the revision I benchmarked by **zero instruction lines** — the only differences are item 3's offsets and item 4's build tag:

```
$ diff <(git show 5a25e04:zstd/seqdec_arm64.s) <(git show HEAD:zstd/seqdec_arm64.s) \
    | grep '^[<>]' | grep -vc '(RSP)\|go:build'
0
```

so the measurement still describes the code in this PR rather than an ancestor of it.

The avo range behind the pin is mostly hardening that produces no visible change here: fourteen rounds of adversarial review found 43 ways the lowering could emit wrong assembly, **none reachable from these generators** — verified by regenerating and getting byte-identical output at each step. The one exception is item 3, where the existing bytes are the bug.

Two of those changes narrow what the lowering accepts, which is worth knowing before it is pointed at anything else: inline `GOAMD64` preprocessor conditionals are now refused outright (that machinery caused 11 of the 43 bugs and no generator here uses it — BMI2 variants go through twin *functions*, which is unaffected), and flag effects are now checked mechanically rather than by inspection.

Happy to split any of the four out if you would rather review them separately.

lizthegrey:
Correcting one thing in my own comment above: I quoted the amd64 `MOVWQZX` win as **−7.59% geomean**, which is literally what benchstat reports but is a misleading way to state it. Updated in the description.

The effect splits cleanly by input size:

| input size | Δ sec/op |
|---|---|
| 100 B (9 cases) | **no significant change**, p = 0.06–0.94 |
| 10 KB and 256 KB (19 cases) | **−6.6% to −14.1%**, all p = 0.000 |

The geomean averages those together, so it simultaneously understates the improvement where the decode loop actually runs and implies a benefit on tiny inputs that is not there.

Two things worth noting about the measurement, since a double-digit gain from 32 instruction substitutions deserves scrutiny:

- The two binaries differed in **exactly 32 lines**, all `MOVW`→`MOVWQZX`. Nothing else was in the delta.
- **The flat 100-byte cases are what rule out a drift artifact.** A thermal or run-ordering effect would have moved them along with everything else; instead they are statistically indistinguishable while the loop-bound cases move 6–14%. Those small inputs are setup-dominated and barely enter the decode loop, which is exactly where the mechanism predicts no effect.

Mechanism, for what it is worth: `MOVW mem, CX` writes only CX's low 16 bits, so the CPU merges with the prior `RCX` — a false dependency on a value the loop immediately consumes for shifts and byte extracts. Removing a loop-carried false dependency from a tight loop is where double digits come from; the `66h` prefix saving is secondary.

lizthegrey:
Rebased onto master now that #1176 is in. Conflict is resolved and CI is running against the new base.

The rebase turned out to be worth more than a conflict fix, because it made the frame fix a **cross-check** rather than a duplicate:

- #1176 fixed the offsets by reasoning about what the lowering should emit, verified by disassembly.
- This branch fixes them *in* the lowering, from the round-seven pin onward.

Those were reached independently, so reproducing one from the other is real evidence. **Regenerating from this pin on top of merged master yields `seqdec_arm64.s` byte for byte** — not a compatible answer, the same one. The round-seven commit here is consequently now assembly-neutral like the other pin bumps, and I reworded it to say so rather than leaving it claiming a `+8` it no longer applies.

Two things this makes easier to review than before:

1. **The 148 lines of `(RSP)` churn are gone**, so `seqdec_arm64.s` now shows only the lowering improvement — 478 diff lines instead of 622. The claim that the benchmarked assembly is unchanged is now checkable in one line: zero instruction-line differences from the benchmarked revision, build tag aside.
2. **The amd64 diff is now entirely the `MOVWQZX` narrowing**, which was previously mixed in with regeneration noise. That is 32 lines, all the same substitution.

Re-validated rather than assumed: full `zstd` and `huff0` suites on Neoverse N1 and amd64, and `go generate` reproducing every checked-in file from a clean tree — the latter being what shows the conflict resolution is the generator's output and not my hand-merge.

klauspost (approved):
Very nice!

(looking to do a bug release before features)

lizthegrey:
Thanks! No rush at all — a bug release first makes sense, and this has no urgency behind it.

It should keep applying cleanly; if it does drift I will rebase and re-run the two checks that matter, namely that `go generate` reproduces every checked-in file from a clean tree, and that `seqdec_arm64.s` comes back byte-identical to the revision the benchmarks were run against.

---

Issue #1114, opened 2025-11-30
flate: Data corruption when compressing with flate.BestCompression


Hakkin:
See https://github.com/kopia/kopia/issues/5049 for initial details.

Reproducer here: https://go.dev/play/p/bU_ax9x8Ct2
It seems to trigger when compressing over a certain amount of data, in this case, just 1MiB of `0x00` results in:
`decompress: flate: corrupt input before offset 10`
when attempted to decompress the just compressed data.

This seems to have been introduced in commit 6e2f5d5b240137060e9d7c44f3c31bad2927593f, released in v1.18.1, commits before that one produce the expected output.

klauspost:
Interesting. It seems to also dependent on the particular alignment of the block size. `TestVeryLongSparseChunk` doesn't seem to hit it.

I will see if I can do a proper fix and avoid the revert. I will probably release and retract the old.

PR #1176, opened 2026-07-25
zstd: fix arm64 asm frame offsets placing locals on the saved LR slot

lizthegrey:
## Summary

The generated arm64 zstd decoder assembly places every function local 8 bytes too low, so **local 0 lands on the saved link register slot**. Present since **v1.19.0**.

**This does not corrupt decoded data and cannot produce wrong output.** Every local shifts by the same amount, so loads and stores agree — which is exactly why the test suite and fuzzing never caught it.

**It also cannot crash any current Go runtime** — the original description of this PR overstated the severity as urgent/crash-capable; see the impact section below for the verified mechanics, and the PR comments for the correction history. The present-day impact is confined to best-effort unwinders: corrupted CPU-profile samples and wrong debugger/core-dump backtraces through these frames on arm64. It is still worth fixing promptly: the frame layout is simply wrong, and it becomes a GC fatal error if the runtime ever starts async-preempting assembly functions (an open TODO in `runtime/preempt.go`).

## Impact (verified against go1.26.5 runtime source and disassembly)

Three mechanisms bound the blast radius:

1. **Normal execution never reads the clobbered slot.** All four functions are leaves, and Go's arm64 leaf epilogue (`ADD $24, RSP, R29; ADD $32, RSP, RSP; RET`) never reloads R30 from the stack — the return address stays live in the register for the whole call. `RET` is unaffected. (If the epilogue did reload LR, this would have been a 100%-reproducible crash on every decode and caught within hours of the v1.19.0 release.)
2. **Precise unwinders can never observe the frame.** GC stack scanning, stack growth, `runtime.Stack`, and goroutine profiles only look at goroutines stopped at safe points, and `isAsyncSafePoint` (`runtime/preempt.go:450`) returns false for any `FuncFlagAsm` function — the runtime spins until the goroutine leaves the assembly. These functions make no calls, so no panic or safepoint can occur inside them either. This is a hard exclusion, not a probability.
3. **The only readers fail soft.** SIGPROF tracebacks land at arbitrary PCs and do read the saved-LR slot (`runtime/traceback.go:372-374` loads it from `frame.sp` for an established innermost frame), as does frame-pointer unwinding (`[R29+8]` is the same slot). Both are best-effort: the garbage PC fails `findfunc` and the sample truncates. No throw.

Net effect today: arm64 CPU profiles of zstd-decode-heavy processes contain samples that truncate at, or show a bogus caller above, `sequenceDecs_*`; out-of-process debuggers and core-dump analysis cannot unwind past these frames. That is the whole present-day impact — and the full explanation of why nothing has been reported in the wild since v1.19.0 shipped.

## Root cause

x86 and arm64 store the return address in different places:

- **x86**: `CALL` pushes it *before* the callee runs, so it sits above the frame and local 0 is at `SP+0`.
- **arm64**: `BL` leaves it in the link register and the callee spills it to the **bottom of its own frame**, so `0(RSP)` is the saved LR and locals begin at `8(RSP)`.

The avo arm64 lowering renamed the pseudo stack pointer to `RSP` and passed avo's (x86-relative) displacement through unchanged, inheriting x86's layout.

## Evidence

Disassembled probe functions at two frame sizes:

```
arm64  TEXT $8-16:   MOVD.W R30, -32(RSP)     <- saved LR at 0(RSP)
                     local-8(SP)  -> 8(RSP)   <- assembler's local 0
                     (RSP)        -> 0(RSP)   <- what we emitted

arm64  TEXT $64-16:  local 0 -> 8(RSP), local 56 -> 64(RSP)
```

The +8 does not depend on frame size. For contrast, on amd64 `TEXT $8-16` emits `SUBQ $8,SP` and `(SP)` really is local 0.

## Affected functions

| function | frame | offsets before | after |
|---|---|---|---|
| `sequenceDecs_decode_arm64` | `$8` | 0 | 8 |
| `sequenceDecs_decode_56_arm64` | `$8` | 0 | 8 |
| `sequenceDecs_decodeSync_arm64` | `$64` | 0–56 | 8–64 |
| `sequenceDecs_decodeSync_safe_arm64` | `$64` | 0–56 | 8–64 |

All post-fix offsets sit inside their frame's local area (`[8, 8+framesize)`).

## The change

Shifts pseudo-SP displacements by 8 in the lowering, so the regenerated assembly is a **mechanical +8 on every local access and nothing else** — reviewable by pattern rather than by reading the whole file. The avo pin moves from `a4dbeac` to `e315f25`, which contains only this fix (branched off the currently-pinned commit).

`amd64` output is unchanged.

### Why this one changes the `.s` when the other lowering work does not

Worth stating explicitly, because it is the opposite of what the companion PR shows. Every other change to the arm64 lowering has been verified by regenerating `zstd` and `huff0` and getting **byte-identical output** — that is the standing check that a printer change cannot affect shipped code.

That check deliberately cannot apply here, because **the current bytes are the bug**. A fix that left `seqdec_arm64.s` unchanged would mean the fix had not taken effect. So the diff is the point, and the way to review it is that its *shape* is uniform:

- 148 changed lines, and **every one carries an `(RSP)` operand** — nothing else in the file moves
- across the four functions, all **78** frame accesses shift by exactly +8
- no instruction added, removed, or reordered; no operand other than the displacement differs
- every resulting offset lands inside its frame's local area, `[8, 8+framesize)`

```
$ git diff master -- zstd/seqdec_arm64.s | grep '^[+-]' | grep -v '^[+-][+-]' | grep -vc '(RSP)'
0
```

One reviewing note, because it caught me: do **not** try to verify this by pairing the `-` and `+` lines in order. git renders the change compactly — a run of stores at `8,16,24` becoming `16,24,32` shows only the removed `8` and the added `32`, with `16` and `24` as unchanged context — so a naive pairing appears to show a +24 jump. Comparing the full offset list per function before and after is the check that actually holds:

```
sequenceDecs_decode_arm64:          3 accesses, every offset +8
sequenceDecs_decode_56_arm64:       3 accesses, every offset +8
sequenceDecs_decodeSync_arm64:     36 accesses, every offset +8
sequenceDecs_decodeSync_safe_arm64: 36 accesses, every offset +8
```

The `generate (zstd)` CI job is the independent confirmation: it regenerates from the pinned avo and fails on any drift, so the checked-in assembly is exactly what `e315f25` produces.

## Testing

- `go test ./zstd/` passes on arm64 (Neoverse N1) and amd64.
- Frame layout verified by disassembly at two frame sizes, as above.
- Impact analysis verified against the go1.26.5 runtime: leaf epilogue disassembly of the shipped functions, `runtime/preempt.go` async-safe-point rules, and `runtime/traceback.go` LR-machine unwind path.

## Relationship to #1175

Found while auditing the avo arm64 lowering, and split out deliberately so it can land on its own — it is a bug fix against released versions, and should not queue behind a larger hardening change.

The two do overlap in one file: #1175 also touches `zstd/seqdec_arm64.s`. **#1175 should rebase onto this**, not the other way round. Its own diff to that file is byte-identical apart from picking up this same fix, so after a rebase the frame change appears once, here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ARM64 decoding reliability by reorganizing internal temporary storage and updating related value handling.
  * Updated the ARM64 decoding dependency to a newer revision for improved compatibility and stability.


lizthegrey:
> [!NOTE]
> **Rescinded 2026-07-29.** The crash scenario this comment assumes cannot occur on current Go runtimes; impact is limited to CPU-profile samples and debugger backtraces. See the [severity correction](https://github.com/klauspost/compress/pull/1176#issuecomment-5123779321). Struck through, retained for history.

~~One release-policy question this raises, which is yours to decide, @klauspost — flagging it rather than assuming either way.~~

~~The bug is in **v1.19.0 and v1.19.1** (absent from v1.18.x), so `retract` is worth considering. I have deliberately **not** added one to this PR, because it does not look like it meets the bar this project has used:~~

| ~~retracted~~ | ~~reason~~ |
|---|---|
| ~~v1.14.1–v1.14.3~~ | ~~data corruption in flate `BestCompression` (#503)~~ |
| ~~v1.18.1~~ | ~~#1114~~ |

~~Both are wrong-output bugs. This one is not: decoded data is correct on every input, which is exactly why the test suite and fuzzing never caught it. The failure is confined to stack unwinding — GC scanning, panics, profiling — on arm64 only, and only while one of the four affected frames is live.~~

~~My read is that a patch release is worth cutting sooner than the next feature release, since `sarama` and `vitess` both pin v1.19.x, but that retraction would be disproportionate and disruptive for a bug that cannot produce wrong output. Happy to add the `retract` block if you disagree — it is a two-line change and I would rather you decide it than inherit my judgement of the severity.~~


**Postscript (unstruck):** the conclusion survives the severity correction — no `retract`, and a patch release ahead of the next feature release is still worthwhile, now for profiling/debugging fidelity rather than crash risk.

lizthegrey:
> [!NOTE]
> **Rescinded 2026-07-29.** The crash scenario this comment assumes cannot occur on current Go runtimes; impact is limited to CPU-profile samples and debugger backtraces. See the [severity correction](https://github.com/klauspost/compress/pull/1176#issuecomment-5123779321). Struck through, retained for history.

### ~~If you arrived here from a crash~~

~~Adding the symptom vocabulary, because the failure does not look like a compress bug from the outside — it looks like a Go runtime bug, and that is where it would naturally be reported.~~

~~A corrupted saved link register only matters when something walks the stack, so the crash surfaces inside the runtime's unwinder rather than in decoding. Plausible shapes, on **linux/arm64 or darwin/arm64 only**:~~

```
fatal error: traceback stuck
runtime: traceback stuck. pc=... sp=...
runtime: unexpected return pc for ...
fatal error: unexpected signal during runtime execution
```

~~with `runtime.gentraceback`, `runtime.(*unwinder).next`, `runtime.tracebackPCs`, `runtime.scanstack`, `runtime.scanobject`, `runtime.sigprof` or `runtime.doRecordGoroutineProfile` on the stack, and `zstd` decoding somewhere in the interrupted goroutine — typically `sequenceDecs_decode_arm64` or `sequenceDecs_decodeSync_arm64`.~~

~~Most likely trigger is a **continuous CPU profiler** (Datadog, Pyroscope, `pprof`), because SIGPROF interrupts arbitrary code and unwinds immediately. GC stack scanning is a less likely trigger, since hand-written assembly is generally not async-preemptible, so the collector will usually not stop a goroutine inside these frames.~~

~~Two caveats, stated plainly: I have **not** reproduced a crash from this — the diagnosis is from the frame layout, which is verified by disassembly. And a `traceback stuck` on arm64 is not automatically this bug; golang/go#62086 collects several unrelated causes, and at least one report there is on a compress version predating this code entirely.~~

~~If you hit something matching the above on v1.19.0 or v1.19.1 with zstd decoding on arm64, a stack dump on this PR would be genuinely useful — it would turn a reasoned diagnosis into a confirmed one.~~

lizthegrey:
> [!NOTE]
> **Rescinded 2026-07-29.** The crash scenario this comment assumes cannot occur on current Go runtimes; impact is limited to CPU-profile samples and debugger backtraces. See the [severity correction](https://github.com/klauspost/compress/pull/1176#issuecomment-5123779321). Struck through, retained for history.

### ~~Exposure is still small, but that depends on release timing~~

~~Audited the significant zstd dependents to see who has actually picked this up. The window is narrow — v1.19.0 is dated 2026-06-30 and v1.19.1 is days old — so **only four projects have shipped it inside a tagged release**:~~

| ~~shipped release~~ | ~~requires~~ |
|---|---|
| ~~`open-telemetry/opentelemetry-collector` v0.157.0~~ | ~~v1.19.0~~ |
| ~~`apache/arrow-go` v18.7.0~~ | ~~v1.19.0~~ |
| ~~`IBM/sarama` v1.60.0~~ | ~~v1.19.0~~ |
| ~~`anchore/syft` v1.49.0~~ | ~~v1.19.0~~ |

~~(Verified by reading `go.mod` at each tag, not inferred.)~~

~~Everything else I found on v1.19.x is **HEAD-only** and would absorb it at its next release: containerd, moby, buildkit, tailscale, talos, soci-snapshotter, VictoriaMetrics, ch-go, google/go-containerregistry, vitess, mimir, iceberg-go, gitea.~~

~~Which is the argument for cutting a patch release sooner rather than folding this into the next feature release: **most of the exposure has not happened yet and a fast v1.19.2 prevents it outright.** The four above are the only ones that would need to re-release.~~

~~Two caveats so this is not read as more precise than it is:~~
- ~~These are *declared* requirements. Minimal version selection can still pull v1.19.x into a build whose own `go.mod` asks for v1.18.x, if any other dependency requires it — I did not compute transitive resolution.~~
- ~~`opentelemetry-collector-contrib` was not enumerated (~200 modules); one sampled component had it indirectly in v0.157.0, so contrib coverage is inferred.~~

~~One useful negative: `nats-io/nats-server` requires v1.19.0 but uses only `s2`, so it is unaffected despite matching on version. Matching the version alone overstates the blast radius — the zstd *decode* path is the thing that matters.~~

lizthegrey:
**Severity correction — this is not urgent, and I have retitled and rewritten the description accordingly.**

The original title/body claimed this could crash any arm64 process that decodes zstd while something walks its stack (GC, panic, profiling). After tracing the actual runtime behavior, that was wrong. Verified against the shipped binaries and the go1.26.5 runtime source:

1. **Normal execution never reads the clobbered slot.** All four functions are leaves; the arm64 leaf epilogue is `ADD $24, RSP, R29; ADD $32, RSP, RSP; RET` — R30 is never reloaded from the stack.
2. **GC scanning, stack growth, `runtime.Stack`, and goroutine profiles can never observe these frames.** `isAsyncSafePoint` (`runtime/preempt.go:450`) refuses to preempt inside any `FuncFlagAsm` function, and the functions make no calls, so precise unwinders only ever run after the frame is gone. Structural exclusion, not luck.
3. **The only readers — SIGPROF tracebacks and frame-pointer unwinds — fail soft.** The garbage "return address" fails symbolization and the sample truncates. No throw.

So the real, present-day impact is: corrupted/truncated CPU-profile samples and broken debugger/core-dump backtraces through `sequenceDecs_*` on arm64. That also fully explains why nothing was reported in the wild since v1.19.0 — a crash was never possible, not merely unlikely. The latent risk that keeps this worth fixing promptly is the TODO directly below `preempt.go:450`: if the runtime ever starts async-preempting assembly, this frame layout becomes a GC fatal error.

The fix itself is unchanged. I have also closed the downstream heads-up issues I filed with the same correction: IBM/sarama#3684, open-telemetry/opentelemetry-collector#15660, apache/arrow-go#1016.

The original framing was out of an abundance of caution — clobbering a saved return address warranted treating as urgent until the runtime mechanics proved otherwise.

klauspost (approved):

klauspost:
ok, good analysis, merging for now. afk for the moment, so will release early next week.

Issue #952, opened 2024-04-18
zstd: Reuse single encoder/decoder with many dictionaries


coxley:
# Problem

I have a multi-tenanted use-case where we will keep hundreds (~500-1000) dictionaries in-memory at any given time. These are used to compress data before writing to storage, gRPC, etc, and decompressed on the way back.

The current API assumes that you know the entire set of dictionaries you'll use at setup time. There's no way to give a dictionary to `w.EncodeAll` nor is there a way to do it for `r.DecodeAll` or even register dictionaries to an existing `*zstd.Reader`.

Ideally, I manage the lifetime of my own dictionaries. When to refresh, prefetch, locally cache, etc. At compression and decompression time, I can handle providing the correct `[]byte` to use. Building a dictionary with `WithEncoderDict` or `WithDecoderDicts` seems to copy all of the data given to it, making additional overhead.

Upstream `zstd` has a function `ZSTD_createCDict_byReference` which avoids copying the input. Doing something similar would be a very nice addition as well.

Any thoughts?

PR #1177, opened 2026-07-25
zstd: don't clear the registered dictionary when decoding past the window

sueun-dev:
While decoding a block whose frame history has grown past the window,
`executeSequences` set `hist.dict.content = nil`. `hist.dict` is the same
`*dict` that `WithDecoderDicts`/`WithDecoderDictRaw` register on the decoder
(`Decoder.setDict` hands the map value straight to `history.setDict`), so this
blanks the registered dictionary for every later decode on that decoder.

After one streaming decode of a dictionary frame whose output exceeds the
window, subsequent decodes of a perfectly valid dictionary frame fail with
`match offset (...) bigger than current history`. That breaks the
"reuse one decoder with many dictionaries" pattern from #952.

The write was also dead for the current decode: the streaming execute path
reads its own `hist.decoders.dict` copy (set in `setDict`), not
`hist.dict.content`, so the assignment only ever corrupted later decodes.
The sync path (`decodeSync`) already keeps the dictionary for the whole decode
and never mutates it.

What changed:
- Drop the `hist.dict.content = nil` block in `executeSequences`; keep the
  `hbytes` window trim. The shared, registered dictionary is left untouched,
  which also removes an unsynchronized write to it during the concurrent
  decode the package documents.
- Add `TestDecoderDictReuseAfterLargeStream`: stream-decodes the same
  dictionary frame three times on one decoder; fails before, passes after.

Checked:
- `go test ./zstd/ -run TestDecoderDictReuseAfterLargeStream` (fails on master, passes with this change)
- `go test ./zstd/` (full package)
- `go test ./zstd/ -run 'Dict|Decoder|RawDict|Reset' -race`
- `go test -fuzz FuzzEncoding -fuzztime 25s` (registers dictionaries; no crashes)
- `gofmt`, `go vet`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where dictionary-based streaming decompression could become invalid after processing large outputs.
  * Dictionary-backed decoding now remains reliable across repeated stream resets.


klauspost:
Please run `go fix`.

klauspost:
@sueun-dev ⬆️

PR #1178, opened 2026-07-25
fix(zstd): handle zero-literal BuildDict corpus

### LINKED ISSUE #1174: zstd: BuildDict panics with "integer divide by zero" on a fully-matchable corpus (litTotal == 0)
### What happened

`zstd.BuildDict` panics with `runtime error: integer divide by zero` when the training corpus is homogeneous enough that the encoder emits zero literals across all samples (`litTotal == 0`).

In `zstd/dict.go`, the literal-table sizing divides by `avgSize`, which is derived from `litTotal` and clamped only on the high side:

```go
// v1.19.1, dict.go:428-431
avgSize := min(litTotal, huff0.BlockSizeMax/2) // == 0 when litTotal == 0
huffBuff := make([]byte, 0, avgSize)
div := max(litTotal/avgSize, 1)                // 0 / 0 panics before max() runs
```

When `litTotal == 0`, `avgSize == 0` and `litTotal/avgSize` is `0/0`. The `max(..., 1)` is meant to guard against `div < 1`, but the division panics before it is evaluated. The `len(contents) == 0` check at dict.go:219 does not cover this case — the content is present, it just produces no literals.

### Reproduce

```go
package main

import (
	"bytes"
	"fmt"

)

func main() {
	block := bytes.Repeat([]byte("ABCDABCDABCDABCD"), 64) // 1 KiB, fully matchable
	var contents [][]byte
	var history []byte
	for i := 0; i < 32; i++ {
		contents = append(contents, block)
		history = append(history, block...)
	}
	_, err := zstd.BuildDict(zstd.BuildDictOptions{
		ID: 1, Contents: contents, History: history, Level: zstd.SpeedDefault,
	})
	fmt.Println("no panic, err =", err)
}
```

Output:

```
panic: runtime error: integer divide by zero

goroutine 1 [running]:
github.com/klauspost/compress/zstd.BuildDict(...)
	.../zstd/dict.go:431
```

### Versions affected

Confirmed on v1.18.0 (panic at dict.go:433) and v1.19.1 (panic at dict.go:431). The refactor to `min`/`max` between those versions did not change the behavior.

### Relation to #950 / #951

This is not #950 (which concerned an RLE-table panic and was closed via #951). PR #951 is where the literal-based sizing was introduced — the `litTotal += len(block.literals)` accumulation and the `avgSize`/`div` computation that this report points at. That fix addressed the RLE case but left the `litTotal == 0` path unguarded, which is what panics here. Reproduces on current `master`/v1.19.1, so it is not fixed by #951.

### Environment

- Go: go1.23 (also reproduced on the local go toolchain)
- OS/arch: darwin/arm64 (the arithmetic is platform-independent)

### Expected

`BuildDict` should return an error (e.g. "not enough literals to train a dictionary") rather than panic, so callers can degrade gracefully instead of guarding every call with `recover()`.

### Real-world trigger

We compress in-memory cache values (HTML/JSON ad payloads) with a per-generation trained dictionary. When the corpus is highly homogeneous (large shared boilerplate repeated across values), the samples are fully covered by matches against the history window, `litTotal` hits 0, and the trainer panics. We currently wrap `BuildDict` in `recover()` as a workaround.

cyphercodes:
## Summary
- return a normal `BuildDict` error when a corpus produces sequences but no literals
- add regression coverage for the fully-matchable corpus panic from #1174

Fixes #1174

## Test plan
- `GOTOOLCHAIN=go1.24.0 go test ./zstd -run 'TestBuildDict(FullyMatchableCorpusReturnsError|LevelPathsRoundTrip)' -count=1`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Dictionary building now stops with a clear error when the input contains no literals, instead of proceeding with empty literal data.
  * Prevents unsafe behavior for fully matchable inputs by ensuring execution fails gracefully rather than panicking.

* **Tests**
  * Added a new test to confirm dictionary building returns an error for fully matchable input and that the error message includes “0 literals”.


klauspost (approved):

klauspost:
@cyphercodes PTAL at `go fix` and conflict. Thanks!

cyphercodes:
Resolved the conflict with current `master` and applied the `go fix` update (`for range 32`).

Verification run locally with Go 1.26.5:
- `diff <(gofmt -d .) <(printf "")`
- `diff <(go fix -diff ./...) <(printf "")`
- `go test ./zstd -run "TestBuildDict(FullyMatchableCorpusReturnsError|HomogeneousCorpusValidOffsets|LevelPathsRoundTrip)$" -count=1`
- `go test ./zstd -count=1`

klauspost:
Thanks. I will see if I can get #1177 in and do a bugfix release before adding features.

PR #1184, opened 2026-08-01
zstd: keep BuildDict recent-offsets positive and loadable

### LINKED ISSUE #1179: zstd: BuildDict can emit a dictionary the library's own encoder rejects with "invalid offset in dictionary"
## Summary

`zstd.BuildDict` can produce a dictionary that the library's own encoder then refuses to load: `zstd.NewWriter(nil, zstd.WithEncoderDict(dict))` returns the error `invalid offset in dictionary`. `BuildDict` returns `err == nil` and a non-empty `[]byte`, but that byte slice is not a usable dictionary — its recent-offsets table contains a non-positive entry, so it is structurally invalid rather than either a valid dictionary or a clean error.

It reproduces on a homogeneous / low-entropy training corpus (many samples sharing a large common prefix plus small variable tails). With the corpus in the repro below it is **deterministic — 100% of `BuildDict` calls produce a rejected dictionary** on both versions tested.

## The error and where it comes from

The error is returned by `loadDict` in `zstd/dict.go` (same location in v1.18.0 and v1.19.1), which validates the dictionary's three recent-offset entries:

```go
// zstd/dict.go, func loadDict
140    if d.offsets[0] <= 0 || d.offsets[1] <= 0 || d.offsets[2] <= 0 {
141        return nil, errors.New("invalid offset in dictionary")   // <-- returned here
142    }
```

`loadDict` is reached from `WithEncoderDict` when constructing the encoder. The offending dictionary is the direct return value of `BuildDict`, so `BuildDict` is emitting a dictionary whose recent-offsets table contains a non-positive entry.

## Reproduction

Self-contained. Each training sample is a fixed boilerplate block shared by every sample plus eight small random-hex payloads; the program builds a dictionary from 16 such samples and feeds it straight to `WithEncoderDict`, 500 times, counting how often the encoder rejects `BuildDict`'s output.

```go
package main

import (
	"fmt"
	"math/rand"

)

// One homogeneous value: a fixed boilerplate block shared by every sample, plus a
// handful of small random-hex payloads.
func sample(seed int) []byte {
	const boiler = `<div class="card-grid" data-template="responsive"><style>.card-grid{display:grid;gap:6px;background:#0a0a0a;padding:4px}.card-item{color:#333;display:grid}</style><script>(function(){var n=document.querySelectorAll('[data-src]');for(var i=0;i<n.length;i++){var u=n[i].getAttribute('data-src');}})();</script>`
	rng := rand.New(rand.NewSource(int64(seed)))
	out := boiler
	for i := 0; i < 8; i++ {
		p := make([]byte, 32)
		rng.Read(p)
		out += fmt.Sprintf(`{"url":"https://example.com/t?id=%d&p=%x"}`, i, p)
	}
	return []byte(out)
}

func main() {
	const sampleCount, historyLimit, N = 16, 8 << 10, 500
	samples := make([][]byte, sampleCount)
	for i := range samples {
		samples[i] = sample(12000 + i)
	}
	var hist []byte
	for _, s := range samples {
		if len(hist)+len(s) > historyLimit {
			break
		}
		hist = append(hist, s...)
	}
	reject := 0
	for n := 0; n < N; n++ {
		dict, err := zstd.BuildDict(zstd.BuildDictOptions{
			ID: 1, Contents: samples, History: hist, Level: zstd.SpeedBetterCompression,
		})
		if err != nil {
			continue
		}
		if _, err := zstd.NewWriter(nil, zstd.WithEncoderDict(dict)); err != nil && err.Error() == "invalid offset in dictionary" {
			reject++
		}
	}
	fmt.Printf("%d BuildDict runs: %d rejected with \"invalid offset in dictionary\" (%.1f%%)\n",
		N, reject, 100*float64(reject)/float64(N))
}
```

Observed output:

```
# github.com/klauspost/compress v1.18.0
500 BuildDict runs: 500 rejected with "invalid offset in dictionary" (100.0%)

# github.com/klauspost/compress v1.19.1
500 BuildDict runs: 500 rejected with "invalid offset in dictionary" (100.0%)
```

The specific rate is corpus-dependent (other homogeneous corpora reject only a fraction of the time — `BuildDict` output is not always deterministic for identical input), but the corpus above triggers it every time.

## Versions affected

Reproduced on both versions tested; the `invalid offset in dictionary` string and the guarding `d.offsets[i] <= 0` check are at `zstd/dict.go:140-141` in both.

| Version | Reproduced | Rate on the corpus above (500 runs) |
|---------|------------|-------------------------------------|
| v1.18.0 | yes        | 100% (500/500)                      |
| v1.19.1 | yes        | 100% (500/500)                      |

Tested with Go 1.23; the behaviour is platform-independent.

## Expected behavior

`BuildDict` must not emit a dictionary that the library's own encoder cannot load. For a given input it should either return a valid dictionary (one that `WithEncoderDict` accepts) or return a clean, non-nil `error` — never `err == nil` together with a byte slice whose recent-offsets table fails `loadDict`'s own `d.offsets[i] > 0` check.

## Relation to #1174

This looks like a separate-but-adjacent robustness gap in `BuildDict` on homogeneous / low-entropy training data. #1174 reports `BuildDict` panicking with an integer divide-by-zero on a zero-literal corpus; this report is about `BuildDict` silently returning a structurally-invalid dictionary (bad recent offsets) that its own encoder later rejects. Both point at `BuildDict` handling highly-redundant training corpora poorly.

zanarellidev:
## Summary

`BuildDict` on a homogeneous corpus with default (zero) `Offsets` could emit a dictionary whose recent-offsets table contained `0`. `WithEncoderDict` then rejected it with `invalid offset in dictionary`.

Root cause: repeat codes (offset 1/2/3) counted `o.Offsets[i]` even when those were still `0`, so `0` ranked as a "most used" offset and was written into the dictionary.

Fix:
- skip zero targets when counting repeat-code hits
- require three positive in-range offsets (pad with `1,4,8` / sequential defaults when ranking alone is short)
- always validate the built blob with `loadDict` before returning

Fixes #1179

## Test plan

- [x] Fail-before / pass-after on `TestBuildDictHomogeneousCorpusValidOffsets` (issue repro corpus)
- [x] `go test ./zstd -run 'TestBuildDict(HomogeneousCorpusValidOffsets|LevelPathsRoundTrip)' -count=1`
- [x] `go test ./zstd -count=1` (full package)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dictionary offset selection for more reliable compression.
  * Prevented invalid or missing offsets from being used.
  * Added validation to catch invalid dictionaries and return clear errors instead of causing crashes.
  * Improved handling of dictionaries trained on uniform HTML and JSON content.

### DIFF
--- a/zstd/dict.go
+++ b/zstd/dict.go
@@ -296,40 +296,81 @@ func BuildDict(o BuildDictOptions) ([]byte, error) {
 			if offset > 3 {
 				newOffsets[offset-3]++
 			} else {
-				newOffsets[uint32(o.Offsets[offset-1])]++
+				// Repeat codes reference the training Offsets. Skip unset
+				// (zero) entries so they are not ranked as real offsets.
+				prev := o.Offsets[offset-1]
+				if prev > 0 {
+					newOffsets[uint32(prev)]++
+				}
 			}
 		}
 	}
 	// Find most used offsets.
 	var sortedOffsets []uint32
 	for k := range newOffsets {
+		if k == 0 {
+			continue
+		}
 		sortedOffsets = append(sortedOffsets, k)
 	}
 	sort.Slice(sortedOffsets, func(i, j int) bool {
 		a, b := sortedOffsets[i], sortedOffsets[j]
-		if a == b {
+		ca, cb := newOffsets[a], newOffsets[b]
+		if ca == cb {
 			// Prefer the longer offset
-			return sortedOffsets[i] > sortedOffsets[j]
+			return a > b
 		}
-		return newOffsets[sortedOffsets[i]] > newOffsets[sortedOffsets[j]]
+		return ca > cb
 	})
-	if len(sortedOffsets) > 3 {
-		if debug {
-			print("Offsets:")
-			for i, v := range sortedOffsets {
-				if i > 20 {
-					break
-				}
-				printf("[%d: %d],", v, newOffsets[v])
+	if debug {
+		print("Offsets:")
+		for i, v := range sortedOffsets {
+			if i > 20 {
+				break
 			}
-			println("")
+			printf("[%d: %d],", v, newOffsets[v])
 		}
-
-		sortedOffsets = sortedOffsets[:3]
+		println("")
 	}
-	for i, v := range sortedOffsets {
-		o.Offsets[i] = int(v)
+	// Dictionary recent-offsets must be three positive values within the
+	// history. Ranked matches may be fewer (or empty when only unset
+	// repeat codes were seen), so fill remaining slots with defaults.
+	used := make(map[int]bool, 3)
+	var finalOffsets [3]int
+	nOff := 0
+	for _, v := range sortedOffsets {
+		iv := int(v)
+		if iv <= 0 || iv > len(hist) || used[iv] {
+			continue
+		}
+		finalOffsets[nOff] = iv
+		used[iv] = true
+		nOff++
+		if nOff == 3 {
+			break
+		}
 	}
+	for _, def := range []int{1, 4, 8} {
+		if nOff == 3 {
+			break
+		}
+		if def <= len(hist) && !used[def] {
+			finalOffsets[nOff] = def
+			used[def] = true
+			nOff++
+		}
+	}
+	for def := 1; nOff < 3 && def <= len(hist); def++ {
+		if !used[def] {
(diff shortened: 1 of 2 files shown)


klauspost (approved):

klauspost:
Please apply `go fix`

Issue #1186, opened 2026-08-03
s2: EncodeBetter assembly is slower than the pure-Go path on incompressible input


lizthegrey:
### What happened

`s2.EncodeBetter` is **slower in assembly than in pure Go** on incompressible input. The Go path is selected by `-tags noasm`, so this compares the two implementations of the same mode.

`BenchmarkRandomEncodeBlock1MB`, amd64 (Ryzen 5 5600X), go1.26, `GOMAXPROCS=1`, `-count=6`, via `benchstat`:

```
                                    │      go      │                 asm                  │
                                    │    sec/op    │    sec/op     vs base                │
RandomEncodeBlock1MB/default          40.34µ ±  2%    24.43µ ± 3%   -39.44% (p=0.002 n=6)
RandomEncodeBlock1MB/better           48.32µ ±  2%    59.48µ ± 3%   +23.10% (p=0.002 n=6)
RandomEncodeBlock1MB/best             2.427m ±  0%    2.367m ± 1%    -2.47% (p=0.002 n=6)
RandomEncodeBlock1MB/snappy-default   40.59µ ±  2%    22.18µ ± 9%   -45.35% (p=0.002 n=6)
RandomEncodeBlock1MB/snappy-better    72.65µ ± 11%    62.48µ ± 8%   -14.00% (p=0.002 n=6)
RandomEncodeBlock1MB/snappy-best      1.440m ±  1%    1.433m ± 1%    -0.52% (p=0.002 n=6)
```

Every other mode is faster in assembly, several by a lot. Only `better` regresses — and notably `snappy-better` does not.

### Likely cause

`s2/_generate/gen.go` clamps the adaptive scan skip at two sites (lines 64 and 81):

```go
o.maxSkip = 100 // Blocks can be long, limit max skipping.
```

With nothing matching, `nextEmit` stays put, so an unclamped `(s-nextEmit)>>skipLog` skip grows geometrically and walks a 1 MB block in ~1.8k steps; clamped at 100 it takes ~10.5k, each doing hash-table lookups that miss cache. That ratio is the right order for the slowdown observed.

I have not confirmed instruction-level attribution for s2 specifically. I did confirm it for the sibling codebase below, where the same clamp produces the same shape and a profile showed ~48% of the encoder sitting on a single scratch-address `ADD` feeding a hash-table load.

### Related

[...]

PR #1187, opened 2026-08-03
s2: arm64 assembly encoders, lowered from the amd64 avo program

lizthegrey:
Adds arm64 assembly for the s2 encoders, which were amd64-only — arm64 fell back to `encode_go.go`. Generated from the **same avo program that produces the amd64 assembly**, lowered to arm64 by an experimental printer in a fork of avo (pinned by commit in `s2/_generate/go.mod`). One source, two architectures.

**The arm64 port leaves amd64 byte-identical to master**, instruction for instruction; only the header comment and build tag differ. A final commit deliberately changes amd64 as well — see [Unconditional TZCNT](#unconditional-tzcnt) — and is separable from the rest.

### Performance, Neoverse-N1 (Ampere), `benchstat` n=6, `GOMAXPROCS=1`

Twain, vs the pure-Go path arm64 used before:

| | 1e5 | 1e6 |
|---|---|---|
| default | **−11.9%** | **−12.4%** |
| snappy-default | **−11.0%** | **−11.9%** |
| better | −5.8% | −8.1% |
| best | ~ | ~ |

Full s2 test suite passes on Neoverse-N1 and Cortex-A72.

### On incompressible input, and #1186

An earlier attempt at this was abandoned partly because `default` regressed ~16% on incompressible input. **That is now fixed** — `default` and `snappy-default` are statistically flat there (p=0.310 / p=0.589), and the `best` modes are unaffected. The intervening work on the lowering (notably folding base+index addressing, which removes an `ADD` per indexed access) closed it.

What remains on incompressible input is `better` at +137.6% and `snappy-better` at +24.6% — and **neither is an arm64 problem**. Both trace entirely to the `maxSkip = 100` clamp, which costs amd64 +23.1% on the same input today; I filed that as #1186. Regenerating arm64 with the clamp removed takes `better` to flat and `snappy-better` to −47.1%, moving the incompressible geomean from +16.6% to −8.7%, with the pure-Go control unmoved.

I am **not** proposing to change the clamp here — as measured in #1186 it is buying real compression ratio on mixed content (up to ~8.7% at 256 KB noise runs), so that is a separate tuning decision. This PR reproduces amd64's behaviour faithfully, including this, on the view that it should be fixed in one place for both architectures rather than diverging arm64 first.

### Unconditional TZCNT

Per @klauspost's review: `TZCNT` needs no `GOAMD64` guard. Its encoding is `BSF`'s with an `F3` prefix, and processors without the instruction ignore the prefix and execute `BSF` — so one encoding is correct everywhere, and the assembler-level choice the generator was making was never needed.

The two differ on a zero input, which cannot reach here (every caller has established the value is nonzero), and in flags, which is where I wanted to be careful before taking it: `TZCNT` takes ZF from the *result* and defines CF, while `BSF` takes ZF from the *source* and leaves CF undefined. For a nonzero input with the low bit set those genuinely disagree. It is unobservable here because **all seventy call sites are followed immediately by `SARQ`**, which overwrites the flags without reading them — verified by scanning the generated output; no conditional jump within two instructions of any of them.

This is the one commit that changes amd64. The diff is exactly the seventy `BSFQ` lines and their surrounding directives; the instruction stream is otherwise identical, modulo avo's operand alignment shifting where the widest opcode in a block changes. arm64 output is byte-unchanged, since it already emitted `TZCNTQ` unconditionally — which is what prompted the question.

**It is the tip of the branch and drops cleanly** if you would rather it landed separately, or after the bug release.

There is a second site the same argument covers that I have left alone: `matchLen`'s `if o.bmi1 { TZCNTQ } else { BSFQ }`. That is a Go-level branch producing separate generated functions rather than an assembler directive, so it is a different mechanism with a wider blast radius. Happy to collapse it too if you want the rule applied uniformly.

### Caveats

- The avo arm64 lowering is **experimental**. It is scalar-only, panics rather than guessing on anything it cannot translate faithfully, and has a differential suite that executes every lowering against a pure-Go reference on both architectures.
- arm64 is emitted by a **separate generator pass**. With the per-site TZCNT directives gone, what remains is the `#ifdef GOAMD64_v4`→`v3` promotion block at the top of the file — an instruction to the amd64 assembler, which arm64 omits. So the passes still have to be separate, though a good deal less is special-cased than before.
- `genMemMoveLong`'s 32-byte loop branches `JA` on the carry a `SUBQ` two instructions earlier left, because x86 `DECQ` deliberately preserves CF where the arm64 `SUBS` it lowers to does not. arm64 gets the same 32 bytes per iteration with the branch reading a `CMPQ` that immediately precedes it. Worth flagging as a latent portability trap in the generator rather than a defect in it.
- `hasAmd64Asm` is renamed `hasAsm`, since arm64 now has assembly too; it still gates the LZ4 converters the same way.

Benchmarks on 4× Neoverse-N1 (Ampere) and 16× Cortex-A72. Glad to rerun anything or test on other cores.

## Summary by CodeRabbit

* **New Features**
  * Added optimized assembly acceleration for ARM64 platforms.
  * Extended compression and decompression paths to use native ARM64 implementations when available.

* **Performance**
  * Improved ARM64 memory-copy and trailing-zero handling for faster encoding and matching.

* **Compatibility**
  * Preserved existing AMD64 behavior while expanding supported architectures.
  * Retained fallback implementations when assembly acceleration is unavailable.


s2/_generate/gen.go
```diff
@@ -21,6 +22,31 @@ import (
...
+	}
+	Comment("#ifdef GOAMD64_v3")
+	// 2016 BMI                 :TZCNT r64, r64                        L:   0.57ns=  2.0c  T:   0.29ns=  1.00c
+	//  315 AMD64               :BSF r64, r64                          L:   0.88ns=  3.1c  T:   0.86ns=  3.00c
+	TZCNTQ(r, r)
+	Comment("#else")
+	BSFQ(r, r)
+	Comment("#endif")
+}
+
```
klauspost, line 55: I've later found out this can unconditionally be [TZCNT](https://www.felixcloutier.com/x86/tzcnt):

> On processors that do not support TZCNT, the instruction byte encoding is executed as BSF.

We do not have cases where 'r' is zero, since we only call this when XOR yields a non-zero value, so the difference is not important.

lizthegrey: Applied in c680ddc — thank you, that is a nicer invariant than the one it replaces.

I checked the two ways it could bite before taking it, since the difference is not only the zero input:

**Zero input** — as you say, unreachable. Every caller has already established the value is nonzero.

**Flags** — these do genuinely differ, and I wanted to be sure nothing read them. `TZCNT` takes ZF from the *result* and defines CF; `BSF` takes ZF from the *source* and leaves CF undefined. So for a nonzero input where the low bit is set, `BSF` leaves ZF=0 and `TZCNT` leaves ZF=1 — a real divergence, just an unobserved one here. Scanning the generated output: 70 sites, and all 70 are followed immediately by `SARQ $0x03`, which overwrites the flags without reading them. No conditional jump within two instructions of any of them.

The resulting amd64 diff is exactly the 70 `BSFQ` lines and their surrounding directives; the instruction stream is otherwise identical modulo avo's operand alignment shifting when the widest opcode in a block changes. arm64 output is byte-unchanged, since it was already emitting `TZCNTQ` unconditionally — which is what prompted the question in the first place.

Two notes:

1. **This does break the "amd64 output is unchanged, byte for byte" property** this PR otherwise holds, so I have kept it as a separate commit at the tip. Drop it and the property comes back, if you would rather this landed on its own or after the bug release.

2. **There is a second site the same argument covers**, which I have left alone: `matchLen`'s `if o.bmi1 { TZCNTQ } else { BSFQ }`. That one is a Go-level branch producing separate generated functions rather than an assembler directive, so it is a different mechanism and a wider blast radius. Happy to collapse it too if you want the rule applied uniformly.

Worth noting for the arm64 half: this removes the *per-site* directives but not the `GOAMD64_v4`→`v3` promotion block at the top of the file, so the arm64 pass still has to be a separate run. It is a good bit less special-cased than before, though.

---

lizthegrey:
@coderabbitai on the duplicate `buildtags.Not("noasm")` in `s2/_generate/gen.go` — the observation is correct, but it is **pre-existing on `master`, not introduced here**:

```
$ git show master:s2/_generate/gen.go | sed -n '/^func main/,/^$/p'
func main() {
	flag.Parse()
	Constraint(buildtags.Not("appengine").ToConstraint())
	Constraint(buildtags.Not("noasm").ToConstraint())
	Constraint(buildtags.Term("gc").ToConstraint())
	Constraint(buildtags.Not("noasm").ToConstraint())
```

This PR adds the `amd64 || arm64` line above them and changes nothing else in that block. Removing the duplicate would rewrite the build-tag line of every generated file in the package, which is churn I would rather not fold into an arm64 port — it makes the diff harder to review for no behavioural change. Happy to send it as a one-line cleanup separately if @klauspost wants it.

---

lizthegrey:
@klauspost nothing further needed from me atm right?

klauspost:
@lizthegrey Just wanted to do a final read-through.

klauspost (approved):

lizthegrey:
> @lizthegrey Just wanted to do a final read-through.

no worries, cheers.

PR #1190, opened 2026-08-06
s2: clamp the Go EncodeBetter skip to match the assembly

### LINKED ISSUE #1186: s2: EncodeBetter assembly is slower than the pure-Go path on incompressible input

lizthegrey:
`encodeBetterBlockAsm` has clamped the adaptive scan skip at 100 since it was written — `o.maxSkip = 100 // Blocks can be long, limit max skipping.` — and `encodeBlockBetterSnappyGo` in this same file already does the same. `encodeBlockBetterGo` did not. After a long stretch with no matches its skip grows geometrically until it strides past the start of the next compressible region and loses the matches there.

This is your suggestion from #1186 — `nextS = s+min((s-nextEmit)>>7+1, maxSkip)` in the Go version above 64K — written to match the existing snappy idiom exactly.

## Measured

4MB of alternating incompressible noise and text, Go before → after:

| noise runs | compressed size | Δ |
|---|---|---|
| 16KB | 2527580 → 2521369 | **−0.25%** |
| 64KB | 2520571 → 2516749 | **−0.15%** |
| 256KB | 2518285 → 2515781 | **−0.10%** |

Uniformly compressible input is **byte-for-byte unchanged** — the skip does not grow while matches keep arriving, so nothing that already compresses well is affected.

The cost lands where you said it was acceptable: on incompressible input, where Go now walks the positions the assembly already walked. `encodeBlockBetterGo` on 1MB of random data goes from ~9µs to ~39µs.

## The outputs are now byte-identical

On the >64KB path the Go and assembly encoders now produce **exactly the same bytes**, at every shape I tried (16KB/64KB/256KB/1MB noise runs in a 4MB block, and plain 64KB+1 and 128KB blocks). The clamp was the only thing they disagreed on there.

The test asserts that equality directly rather than bounding a ratio gap, since it is a stronger and more informative check. Verified to fail without the clamp — by +0.25%, +0.15%, +0.10% and +3.17% at the four run lengths.

## One thing I did not change

Blocks of **16383..64KB** still diverge, in the other direction. The assembly dispatches `encodeBetterBlockAsm4MB` from `limit12B` up, which clamps; Go dispatches `encodeBlockBetterGo64K`, which does not — and also uses `skipLog` 6 rather than 7, so it is a different configuration rather than a missing clamp.

On uniform inputs the two still agree byte for byte at those sizes. They part company only on mixed content with long noise runs inside a small block, where I measured Go64K emitting 4%–16.5% more than the assembly. Since you framed the fix as "above 64K" and this would need a new dispatch tier rather than a one-line clamp, I have left it alone — happy to look at it separately if it is worth having.

Full s2 suite passes.

Fixes #1186

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compression match searching by limiting adaptive skip distances to 100 bytes, helping prevent inefficient searches on certain inputs.
  - Improved consistency between compression paths for large data blocks, ensuring equivalent encoded results.

- **Tests**
  - Added coverage for varied compressible and incompressible data patterns and run lengths.
  - Verified matching encoded output sizes and bytes for inputs larger than 64 KiB.


s2/encode_better.go
```diff
@@ -86,12 +86,18 @@ func encodeBlockBetterGo(dst, src []byte) (d int) {
...
+	// compressible region and loses the matches there.
+	const maxSkip = 100
+
 	for {
 		candidateL := 0
 		nextS := 0
 		for {
 			// Next src position to check
-			nextS = s + (s-nextEmit)>>7 + 1
+			nextS = min(s+(s-nextEmit)>>7+1, s+maxSkip)
```
klauspost:
```suggestion
			nextS = s+min((s-nextEmit)>>7+1, maxSkip)
```

This should make it a constant compare and has one less addition. There is a chance it is faster on some platforms.

klauspost (changes requested):

s2/encode_better.go
klauspost: @lizthegrey Sorry, handtyped it. It seems to need formatting :)

lizthegrey: Applied, gofmt'd — `nextS = s + min((s-nextEmit)>>7+1, maxSkip)`.

Agreed on the reasoning, and it is also the safer form: hoisting `s` out means the comparison cannot overflow on a 32-bit `int`, where the old shape adds `s` to both sides before comparing. Not reachable at these block sizes, but free to avoid.

Re-verified after the change:

- Go and assembly output still **byte-identical** on the >64KB path at all four shapes.
- Mutation check still fails without the clamp (+0.25% / +0.15% / +0.10% / +3.17%).
- Full s2 suite passes.

One question while you are here: the two snappy sites in this same file (lines 361 and 781) still use the old shape:

```go
nextS = min(s+(s-nextEmit)>>7+1, s+maxSkip)
nextS = min(s+(s-nextEmit)>>6+1, s+maxSkip)
```

Same argument applies to both, and it would leave the file consistent rather than one-of-three converted. I have left them alone since you commented only on the line this PR adds — say the word and I will fold them in, or send them separately.

---

lizthegrey: Apologies for the force-push churn on the timeline — I first applied this as an amend, which would have left you without a "changes since your review" diff. Re-pushed so the commit you reviewed (`d899da4`) is intact and the change is a separate commit on top (`ee01404`), which is what it should have been in the first place. No further rewrites.

---

klauspost (approved):

PR #1192, opened 2026-08-11
gzhttp: limit jitter hash to jitterBuffer bytes

### LINKED ISSUE #1191: gzhttp: jitterBuffer logic hashes entire buffer indiscriminately
In `gzhttp/compress.go` the middleware supports adding a random jitter to the compressed output. The user can specify a `jitterBuffer` to ensure the jitter is calculated deterministically from the first N bytes of the response so the random jitter does not change based on how the application chunks its `Write()` calls.

The comment explicitly states:
`// Use only up to "w.jitterBuffer", otherwise the output depends on write sizes.`

However the implementation fails to enforce this. When a large initial `Write()` occurs the internal buffer `w.buf` fills up to `wantBuf` (which defaults to `max(w.minSize, 512)`). Then `startCompression` hashes the entire `w.buf` without slicing it down to `w.jitterBuffer`:

```go
if w.sha256Jitter {
    h := sha256.New()
    h.Write(w.buf) // BUG: hashes the entire w.buf which could be up to wantBuf bytes
    // Use only up to "w.jitterBuffer", otherwise the output depends on write sizes.
    if len(remain) > 0 && len(w.buf) < w.jitterBuffer {
        // ...
    }
} else {
    h := crc32.Update(0, castagnoliTable, w.buf) // BUG: same issue here
    // ...
}
```

Because `h.Write(w.buf)` writes the entire buffer indiscriminately the jitter RNG seed becomes highly dependent on whether the first `Write()` call was small or large. This defeats the purpose of the `jitterBuffer` configuration. 

To fix this it needs to be constrained like so:
```go
hashSize := len(w.buf)
if w.jitterBuffer > 0 && hashSize > w.jitterBuffer {
    hashSize = w.jitterBuffer
}
h.Write(w.buf[:hashSize])
```

VedantMadane:
## Summary

Fixes #1191.

When `RandomJitter` is configured with a `jitterBuffer` smaller than the internal write buffer (`wantBuf = max(minSize, 512)`), `startCompression` hashed the **entire** `w.buf` for the padding RNG seed. That made the seed depend on how the application chunked its `Write()` calls, which is exactly what `jitterBuffer` is documented to avoid.

### Fix

Clamp the hashed prefix to `jitterBuffer` for both CRC32 and SHA-256 (paranoid) paths, and apply the same limit when topping up from `remain`:

```go
hashBuf := w.buf
if len(hashBuf) > w.jitterBuffer {
    hashBuf = hashBuf[:w.jitterBuffer]
}
```

### Test

Adds `TestRandomJitterBufferWriteSizeIndependent`, which asserts that for `jitterBuffer=64` and `MinSize(10)`:

- one-shot `Write`, byte-at-a-time writes, and a large first chunk all produce the same compressed length
- mutating bytes **after** `jitterBuffer` does not change padding length
- mutating bytes **within** `jitterBuffer` does change padding length

Covers both CRC32 and paranoid (SHA-256) modes.

## Test plan

- [x] `go test ./gzhttp -run 'TestRandomJitter|TestRandomJitterBufferWriteSizeIndependent|TestRandomJitterParanoid' -count=1`
- [x] `go test ./gzhttp -count=1` (full package)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved random padding consistency regardless of how data is written or buffered.
  * Ensured data beyond the configured jitter buffer does not affect padding calculations.
  * Preserved expected behavior for both standard and paranoid modes.

* **Tests**
  * Added coverage for one-shot, byte-wise, and large-chunk writes.
  * Verified padding responds only to relevant data within the jitter buffer.


klauspost (approved):

klauspost:
Seems a `go fix` is needed.

VedantMadane:
Automated fix: apply go fix range-over-int for build-special CI.

PR #1197, opened 2026-08-29
zstd/huff0: break false partial-register dependencies in seqdec/4X decoders

lizthegrey:
## Summary

Two independent generator-level fixes that remove false/partial-register write-after-write dependencies in the hand-written decode assembly:

- **586af67** — huff0: the 4X decoders wrote each decoded symbol via `MOVB CH,AL` / `MOVB CH,AH` into a fixed `AX`, then stored `AX` as a word. The `AL` write is an x86 partial-register **merge** (post-Sandy Bridge CPUs insert a merge uop that depends on `AX`'s previous value), so all four nominally-independent bitstreams serialize through one register across loop iterations. On arm64 the lowering emits `BFI`, a true read-modify-write. Fix: write the first symbol via `MOVBLZX`/`MOVBLZX`-equivalent (a fresh full-register def), so the second byte's write depends only on same-iteration state. 8 sites across the 4X and 4X_8b loops (generator-level, `huff0/_generate/gen.go`).
- **66ea608** — zstd: the non-BMI2 seqdec had `MOVB AH,CL` / `MOVB nBits,CL` sites (24 total across `decode`, `decode_56`, `decodeSync`, `decodeSync_safe`) that were correct only because `brBitsRead ≤ 64` leaves the upper bits of the destination register zero — a latent known-bits assumption. Fix: `MOVZBL`, which is unconditionally correct (drops the known-bits assumption) and collapses the arm64 `UBFX+BFI` RMW pair to a single fresh `UBFX`/`MOVBU`. **The BMI2 amd64 variants are byte-identical before and after** — this commit only changes the non-BMI2 amd64 path and the arm64 lowering.
- **99ef599** — zstd: the same false-dependency shape, one more site: `copy_slow_3`'s byte-at-a-time overlapping-match copy loop read each source byte via `MOVB` into a `GP64` declared once outside the loop, so the partial-register write merged with the register's value from the *previous* iteration, serializing an otherwise-independent byte copy. Found via a second pass over the same production profile after the first two fixes shipped. Fix: `MOVBQZX`, a fresh full-register def (identical amd64 cost); on arm64 this also drops a now-unneeded `BFI` merge entirely.

Motivation: production `pprof -list` from a Go/arm64 service (two host groups, ~9.6 and ~12.7 CPU-hours) showed `zstd.sequenceDecs_decodeSync_arm64` at 24.1%/18.5% flat and `huff0.decompress4x_main_loop_arm64` at 11.6%/8.6% flat, with samples concentrated on the `UBFX`/`BFI` pairs (99ef599's `copy_slow_3` site carried ~0.7–0.9% of `decodeSync`'s flat time in the same profile).

## Benchmarks

Methodology: `benchstat`, n=10, `-cpu=1`, pinned core, master/branch rounds alternated to cancel drift. Assembly changes shift the byte offset of everything after them in the same `.s` file, which can flip an unrelated (byte-identical) function's cache-line alignment class and produce a phantom multi-percent delta — confirmed via `objdump`/`nm` on both amd64 (BMI2 twins, `executeSimple`) and arm64 (`executeSimple_arm64`). All numbers below are from binaries built with a temporary `PCALIGN $64` inserted before every `TEXT` symbol in the affected `.s` files (measurement-only; not part of the diff), which pins every function to the same alignment class in both binaries and eliminates that artifact. Untouched functions (`executeSimple`, huff0 1X) are included as in-situ controls and sit flat everywhere below, confirming the correction worked.

### x86 (AMD Ryzen 5 5600X / Zen 3, BMI2-capable)

| Path | geomean | note |
|---|---|---|
| zstd non-BMI2 `decode*` (actually changed) | **−1.04%** | modest, consistent, no regressions |
| zstd BMI2 `decode`/`decodeSync`/`execute` | ~0% | expected — byte-identical before/after |
| huff0 4X (`Decompress4X*`) | **−2.6% to −5.5%** per input, no regressions | |
| huff0 1X (control, untouched) | ~0% | |

### arm64

| Machine | zstd `decode` (touched) | zstd `decodeSync` (touched) | zstd `execute` (control) | huff0 4X (touched) | huff0 1X (control) |
|---|---|---|---|---|---|
| Oracle Cloud Ampere Neoverse-N1 (Graviton2-class) | −2.93% | −0.18% | −0.07% | −2.04% | −0.01% |
| AWS Graviton3 (c7g.xlarge, Neoverse-V1) | **−13.23%** | −1.99% | −0.09% | −1.24% | 0.00% |
| AWS Graviton4 (c8g.xlarge) | **−10.28%** | −3.17% | −1.61% (n.s., p=0.14) | −1.72% | 0.00% |
| AWS Graviton5 (c9g.xlarge) | **−13.32%** | −2.96% | +0.03% | −4.11% | +0.01% |

All 14 block sizes in `Benchmark_seqdec_decode` are individually significant (p < 0.001) and directionally uniform on every machine — not outlier-driven. The win is *larger* on the newer, wider Graviton3/4/5 cores than on N1: consistent with a fixed-cycle dependency stall becoming a bigger share of a shorter critical path as the surrounding code gets faster on wider out-of-order cores. No regressions observed on any of the five machines tested.

### 99ef599 (`copy_slow_3`), Oracle Cloud Ampere Neoverse-N1

`BenchmarkDecoder_DecodeAll`, `benchstat` n=10 (5 alternated rounds), `PCALIGN`-pinned as above: geomean **−0.98% sec/op (~+1.0% throughput)**, zero regressions. `html_x_4.zst` (heavy on long-distance overlapping matches, i.e. exactly the `copy_slow_3` path) sees the largest single-file win at **−7.15%**.

## Validation

- All three commits: `go vet` + build on amd64/arm64; existing zstd/huff0 test suites (asm and `-tags=noasm`); `FuzzCompress` and `FuzzDecodeAll` (low/high-mem differential) fuzzing.
- 66ea608 additionally covered by `FuzzDecAllNoBMI2`, which disables BMI2 at runtime and byte-compares the changed non-BMI2 output against the unchanged BMI2 output on the same fuzz corpus — a strong oracle for this specific class of change, since it's exactly a non-BMI2-vs-BMI2 differential.
- Diff shape is fully generator-output: no hand-edited `.s`, no avo-lowering-printer changes. Regenerating both packages produces byte-identical output to what's committed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Huffman and sequence decompression reliability by ensuring decoded values are fully initialized before use.
  * Resolved potential issues caused by stale register data in optimized decompression paths.
  * Improved consistency across standard, synchronized, and safe sequence-decoding paths.
  * Applied fixes across supported amd64 and ARM64 implementations without changing external behavior.


klauspost (approved):
Nice find!

lizthegrey:
> Nice find!

Slowly working through all the profiling findings! it's super nice it's *both* a win on x86 and on arm!

PR #1198, opened 2026-08-29
ci: test against go1.27.x, drop go1.24.x

lizthegrey:
go1.27.0 has shipped stable (confirmed against go.dev/dl). Per the two-latest-release support policy, the currently supported set is 1.26/1.27, so this drops `1.24.x` from the CI build matrix and adds `1.27.x`.

`go.mod`'s `go 1.24` minimum-version directive is left untouched — that's a separate downstream-compatibility decision from which versions CI exercises, and out of scope here.

## Summary by CodeRabbit

* **Compatibility**
  * Updated the minimum supported Go version to 1.25.
  * Added build coverage for Go 1.25.x, 1.26.x, and 1.27.x.
  * Removed build coverage for Go 1.24.x.

* **Tests**
  * Improved compression examples to validate decompressed round-trip output.
  * Added error checks for compression stream processing and closure.
  * Updated concurrency tests and examples for newer Go synchronization features.


lizthegrey:
Adding go1.27.x surfaced a real (if narrow) test issue: \`zlib/example_test.go\`'s \`ExampleNewWriter\` imports Go's **standard library** \`compress/zlib\` (not this package — a vestige of this repo's origin as a zlib/flate fork) and asserted an exact golden byte sequence for the stdlib flate encoder's output. Go 1.27 changed that encoder's tiny-input heuristic (now stores 13-byte inputs instead of Huffman-compressing them — a legitimate stdlib change; exact encoder output was never part of Go's compatibility promise, only decodability is). Since the matrix now spans 1.25/1.26/1.27, no single golden byte sequence can be correct for all three, so I rewrote the test to round-trip through a Reader instead, matching the existing style of \`ExampleNewReader\` in the same file. Verified locally against both go1.26.7 and go1.27.0, plus the full \`go test ./...\` suite on both.

(Checked \`gzip/example_test.go\` for the same latent pattern — it already round-trips rather than asserting raw compressed bytes, so no change needed there.)

lizthegrey:
Addressed both CodeRabbit findings:
- Round-trip errors in \`ExampleNewWriter\` are now checked — matters here since the zlib checksum only verifies at \`Close()\`, so an unchecked error could silently mask corruption that \`io.Copy\` had already written correct-looking bytes for.
- \`go.mod\`'s minimum bumped from \`go 1.24\` to \`go 1.25\` to match what CI now actually covers, rather than leaving an untested compatibility claim. This is a breaking change for consumers still on Go 1.24 (the \`go\` directive is a hard toolchain minimum since Go 1.21) — flagging explicitly since it's a real compatibility decision, not just a CI tweak.

klauspost (approved):

PR #1199, opened 2026-08-30
zstd/huff0: fold two ISA-gap x86 idioms in the avo arm64 lowering

lizthegrey:
## Summary

Two arm64-only lowering peepholes in the [honeycombio/avo](https://github.com/honeycombio/avo) fork's printer (`printer/arm64.go`, pinned via `honeycombio/avo@0a52965` in `zstd/_generate/go.mod` and `huff0/_generate/go.mod`), regenerated here. Both fold a multi-instruction x86 idiom -- needed only because the amd64 ISA lacks a single instruction for the operation -- into one arm64 instruction that has no such gap. **amd64 output is byte-identical**; only the arm64 `.s` files change.

- **TESTB/TESTW `dst, dst`** (x86's "is dst zero?" idiom, lowered today as two independent zero-extended scratch copies + `TST`) now folds to a single immediate-masked `TST` against the original register -- sound unconditionally, no liveness scan needed, since it's a same-register special case of the printer's existing `lowerSubwordTestEqNe`. Hits the loop-bottom exhausted-flag check in huff0's 4X decode loops and the FSE table-build check (3 instructions -> 1, three sites).
- **`MOVQ/MOVL src, tmp; SHRQ/SHRL $n, tmp; MOVBQZX/MOVBLZX tmp, tmp`** (x86's two-address-register idiom for extracting one byte at a fixed bit offset) folds to a single `UBFX`. Restricted to the self-referential form (the extend writes back into the exact register the shift shifted), which makes it provably safe with no forward liveness proof -- nothing in the strictly-adjacent three-instruction window can read the intermediate value, and nothing after can observe it either, since the fold reproduces the window's final register state exactly. Hits the "moB" byte extract in all four zstd `sequenceDecs_*` functions (4 instructions -> 1, also dropping a dead `MOVD` to the stack).

Motivation: production `pprof -list` from Honeycomb's arm64 Lambda workload (Graviton2) flagged both instruction shapes as hot in the same functions #1197 already touched. Full candidate list (5 shapes from the profile) was reviewed by a Fable subagent for correctness before building anything; three were rejected or deferred -- one (a shift-count-materialization fold in huff0's `peekTopBits`) is the exact peephole hand-spiked and killed in an earlier session (ship bar violated: 1X flat, 4X only hit the low end with 5 cells regressing >1%); two others need a liveness scan this pass doesn't attempt and were left for a follow-up. Only the two correctness-clean, no-liveness-required folds above are in this PR.

## Correctness

- New printer logic lives behind `shiftExtractFold`/`emitShiftExtractFold` and a same-register fast path in `lowerSubwordTestEqNe` (avo `printer/arm64.go`).
- New differential tests in avo's `tests/arm64lower` property harness (`ShiftExtractByte64Lo/Mid/Hi`, `ShiftExtractByte32`, `TestBSelfZero`, `TestWSelfZero`), run against a pure-Go reference on both amd64 and real arm64 hardware (Ampere Neoverse-N1).
- Both folds were **mutation-checked**: a deliberately wrong mask and a deliberately wrong shift amount were each confirmed to make the corresponding new test fail (the shift-amount mutation failed loudly enough to be an assembler rejection), confirming the tests exercise the new code paths rather than passing vacuously.
- Regenerating zstd/huff0 from the pinned avo commit changes exactly the matched sites and nothing else -- verified by diff.
- Full `go vet` + build on amd64/arm64; full zstd/huff0 test suites (asm and `-tags=noasm`) passing on real arm64 hardware (Oracle Cloud Ampere N1).

## Benchmarks

Methodology: `benchstat`, n=10 (5 alternated before/after rounds, `-cpu=1`), on Oracle Cloud Ampere Neoverse-N1 (Graviton2-class). Binaries built from `PCALIGN $64`-pinned copies of the affected `.s` files (measurement-only, not part of the diff) to cancel code-alignment artifacts from the instruction-count change shifting downstream function offsets.

| Benchmark | geomean | regressed cells |
|---|---|---|
| zstd `BenchmarkDecoder_DecodeAll` (all 12 corpus files) | **+0.80% throughput** (−0.79% sec/op) | 0 -- every file improves except one flat (execute-bound outlier, consistent with prior PRs) |
| huff0 `BenchmarkDecompress4X*` (all variants/inputs) | **+0.68% throughput** | 0 -- smallest deltas are flat, nothing regresses |

Both wins are modest but statistically robust (p < 0.001 on nearly every cell) and carry zero downside cells, unlike the killed candidate mentioned above. Smaller than #1197's numbers, as expected -- these folds each remove 1-3 instructions at one site per iteration, versus #1197's fix to a loop-carried dependency chain.

## Validation

- `go vet` + build on amd64/arm64; zstd/huff0 test suites (asm and `-tags=noasm`) on real arm64 hardware.
- Diff shape is fully generator-output: no hand-edited `.s` in this package, only the avo printer and the pinned commit. Regenerating zstd/huff0 produces byte-identical output to what's committed.
- amd64 output confirmed byte-identical before/after (both folds are arm64-only printer changes).

## Summary by CodeRabbit

* **Performance**
  * Improved ARM64 decoding efficiency through streamlined low-byte checks and offset-state extraction.
  * Applied optimizations across multiple Huffman, FSE, and sequence decoding paths.
* **Maintenance**
  * Updated code-generation tooling references to newer versions.


klauspost (approved):

huff0/_generate/go.mod
```diff
@@ -17,4 +17,4 @@ require (
 
 replace github.com/klauspost/compress => ../..
 
-replace github.com/mmcloughlin/avo => github.com/honeycombio/avo v0.6.1-0.20260725101136-a995462d70fa
+replace github.com/mmcloughlin/avo => github.com/honeycombio/avo v0.6.1-0.20260830035156-0a52965da4bd
```
klauspost, line 20: @lizthegrey As a sidenote, I don't know if @mmcloughlin has the time or interest to handle this feature.

Maybe consider doing it as a "full-fledged" fork, so we can replace the actual imports, you can do releases, etc.

That also gives other users a place to report issues, contribute, etc. It can ofc still be merged upstream.

lizthegrey: I've spoken to @mmcloughlin recently and volunteered to join as a co-maintainer if he is interested.

PR #1202, opened 2026-09-02
Preallocate buffers to avoid multiple allocations

GavinDmello:
I noticed multiple preallocations while decompressing large protobuf objects. Using this preallocation strategy, we can reduce calls to runtime.growSlice. 

## Summary by CodeRabbit

- **Performance**
  - Improved decompression performance for large sequence blocks by preallocating output capacity more efficiently.
  - Reduced incremental buffer growth during sequence decoding.


lizthegrey:
> I noticed multiple preallocations while decompressing large protobuf objects. Using this preallocation strategy, we can reduce calls to runtime.growSlice.

do you have any benchmarks?

zstd/seqdec.go
```diff
@@ -233,6 +233,27 @@ func (s *sequenceDecs) decodeSync(hist []byte) error {
 	out := s.out
 	maxBlockSize := min(s.windowSize, maxCompressedBlockSize)
 
+	var estimatedSize int
+	if seqs > 0 {
+		avgLitPerSeq := len(s.literals) / seqs
```
klauspost, line 238: Very likely to round down to zero.

zstd/seqdec.go
```diff
@@ -233,6 +233,27 @@ func (s *sequenceDecs) decodeSync(hist []byte) error {
...
+		avgBytesPerSeq := avgLitPerSeq * 3
```
klauspost, line 239: Seems completely arbitrary to me.

zstd/seqdec.go
```diff
@@ -233,6 +233,27 @@ func (s *sequenceDecs) decodeSync(hist []byte) error {
...
+		estimatedSize = len(s.literals) + (seqs * avgBytesPerSeq)
+	}
+
+	// Preallocate buffer to avoid multiple allocations done by runtime.growSlice
+	if estimatedSize > maxBlockSize {
```
klauspost, line 244: `maxBlockSize` is a hard limit, a block will never be bigger than that.

klauspost (changes requested):
This feels too much like a "vibe pr" to be honest, changing random stuff until some random benchmark shows a better number.

There are plenty of ways to avoid append allocs:

A) Use [WithDecoderLowmem(false)](https://pkg.go.dev/github.com/klauspost/compress@v1.20.0/zstd#WithDecoderLowmem). It will preallocate for bigger blocks.
B) Use  [WithSingleSegment](https://pkg.go.dev/github.com/klauspost/compress@v1.20.0/zstd#WithSingleSegment) or [ResetContentSize](https://pkg.go.dev/github.com/klauspost/compress@v1.20.0/zstd#Encoder.ResetContentSize) when encoding. This will record the uncompressed size so it can be allocated up front.
C) Provide a "big enough" buffer when decompressing.

GavinDmello:
> There are plenty of ways to avoid append allocs:
A) Use [WithDecoderLowmem(false)](https://pkg.go.dev/github.com/klauspost/compress@v1.20.0/zstd#WithDecoderLowmem). It will preallocate for bigger blocks.
B) Use [WithSingleSegment](https://pkg.go.dev/github.com/klauspost/compress@v1.20.0/zstd#WithSingleSegment) or [ResetContentSize](https://pkg.go.dev/github.com/klauspost/compress@v1.20.0/zstd#Encoder.ResetContentSize) when encoding. This will record the uncompressed size so it can be allocated up front.
C) Provide a "big enough" buffer when decompressing. 

@klauspost Thanks for the feedback! I'll explore these options.

PR #1204, opened 2026-09-04
zstd: copy the first block of short literal runs and matches unconditionally

lizthegrey:
Follow-up to #1203.

## What

In the extended (unsafe) copy variants of the sequence decoders, the literal copy and the non-overlapping match copy now write their first 16-byte block unconditionally and branch only for the tail beyond 16 bytes. The safe variants and every allocation contract are unchanged: the literal buffer already has to be readable 16 bytes past its end and the output already has to carry 16 bytes of slack past `ll+ml` for these variants to be selected.

## Why

Measured on `testdata/benchdecoder.zip` (Go reference path): 68% of sequences have no literals, 99% have at most 16, 92% of matches are at most 16 bytes, and 99% of offsets are at least 16. The `ll == 0` test was therefore a two-in-three data-dependent branch, and both block loops exited after one iteration nearly every time. The production arm64 profile that motivated #1203 put the literal copy, match copy loop, and the loop handoff at about a quarter of `decodeSync`.

## Numbers

DecodeAll, `benchdecoder` text files, default level, PCALIGN-pinned per-function alignment, sec/op:

| | Tom Sawyer | pi | e | html | other files |
|---|---|---|---|---|---|
| Neoverse N1 (oracle, -count 8) | -10.7% | -7.2% | -6.9% | -8.0% | within ±0.7% |
| Sapphire Rapids (c7i, -count 10 -cpu 1) | -23.3% | -8.8% | -8.7% | +1.0% | within ±0.8% |
| Granite Rapids (c8i, -count 10 -cpu 1) | -20.9% | -8.4% | -8.3% | -1.3% | within ±0.7% |

`Benchmark_seqdec_decodeSync` large blocks are within ±1% except one at -11% (N1) / -17% (both Xeons). Those blocks are literal-heavy (a literal run on almost every sequence), so the zero test was already predictable there. Its small blocks never reach the assembly (the benchmark allocates no output buffer, so `decodeSyncSimple` declines and the Go loop runs) and only move by code layout; I'll fix that benchmark separately.

## Validation

- `go test ./zstd/ ./huff0/` passes on amd64 (Zen 3, Sapphire Rapids, Granite Rapids) and arm64 (Neoverse N1); `GOARCH=arm64 go vet ./zstd/` clean.
- New `TestDecoderShortSequenceCopies` round-trips inputs whose literal runs and matches sit at 0, 15, 16 and 17 bytes through both the DecodeAll and streaming paths. Mutation check: with the match tail threshold off by one it fails with a CRC error.

## Summary by CodeRabbit

- **Performance**
  - Improved decompression efficiency for short literal and match sequences on supported amd64 and arm64 systems.
  - Reduced overhead in fast decoding paths, including cases with empty literal runs or matches of 16 bytes or fewer.

- **Reliability**
  - Added randomized coverage for short-sequence boundary cases across full-buffer and streaming decoding.


zstd/seqdec_copy_test.go
```diff
@@ -0,0 +1,93 @@
...
+		for _, level := range levels {
+			t.Run(tc.name+"/"+level.String(), func(t *testing.T) {
+				enc, err := NewWriter(nil, WithEncoderLevel(level))
+				if err != nil {
+					t.Fatal(err)
+				}
+				compressed := enc.EncodeAll(input, nil)
+				enc.Close()
+
+				got, err := dec.DecodeAll(compressed, nil)
```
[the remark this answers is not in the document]
lizthegrey: Partly addressed in b9d51ce. The names did over-claim: the encoder decides the sequences, and the inputs only steer them to cluster at the boundary. The cases are now named by their construction (word length and literal-run length) and the doc comment says what is and is not guaranteed.

On proving the path: `DecodeAll` allocates its buffers with `compressedBlockOverAlloc` slack by construction (decoder.go / framedec.go), and its `sequenceDecs` is not reachable from a test, so routing this test through `useSafeDecodeSync()` is not possible without exporting internals. The selector is asserted directly on recorded blocks in #1205, which now runs each block through the Go, safe and unsafe implementations and checks `useSafeDecodeSync()` before each. On coverage of the boundary lengths: this test was mutation-checked against the generator, with the match tail threshold off by one (17 instead of 16) the 17-byte-word cases fail with a CRC error, which is direct evidence that sequences at the boundary are produced and go through the extended copies.

zstd/_generate/gen.go
```diff
@@ -1195,13 +1196,22 @@ func (e executeSimple) executeSingleTriple(c *executeSingleTripleContext, handle
...
+			// literals and 99% have at most 16, so the zero test is a
+			// poorly predicted branch and the block loop nearly always runs
+			// exactly once. Copy the first block unconditionally -- an
+			// empty run copies 16 bytes of slack that the match copy then
+			// overwrites -- and loop only for the long tail. The literal
+			// buffer has compressedBlockOverAlloc readable bytes past its
+			// end and the output has the same slack past ll+ml (see
+			// useSafeDecodeSync / executeSimple), which covers the ll == 0
+			// read and write.
+			e.copyMemoryND16("1", c.literals, c.outBase, ll)
```
klauspost, line 1214: Reasonable, but maybe test 8 bytes unconditionally. 

My experience with stats tell me that there is often a huge dropoff at 8 literals already on most inputs.

lizthegrey: I'll have a look

klauspost: (it will be a tradeoff between slightly worse prediction and over-copying, so if you have it available check both platforms)

16 will probably be fine, just want to check.

lizthegrey: If I interpreted your request correctly, you want one 8-byte GPR move, and only branch to the block loop above 8 bytes. I measured that against the current 16-byte first block on Zen 3 and Neoverse N1, PCALIGN-pinned. Where the drop-off at 8 holds (Tom Sawyer: 1.6% of literal runs exceed 8) it is exactly flat, since the 8-byte and 16-byte moves cost the same on both cores. Where it does not (e.txt and pi.txt: 15% exceed 8 versus 4% exceed 16) it is 3-4.5% slower on both, because the tail branch goes from one-in-25 to one-in-7 and stops predicting. So 16 covers the 9-16 band for free and I would keep it, unless you meant something else.

Edit: Counters on the N1 back that up: with the 8-byte first block e.txt retires +0.4% instructions but +73% branch misses (6.4M to 11.1M per 2000 decodes) for +3.1% cycles, about 20 cycles per extra miss, so the over-copying is free and the whole cost is prediction. Tom Sawyer shows the same at a tenth of the scale. Keeping 16.

klauspost: It's fine. `e.txt` and `pi.txt` are not really representative, but neither is Tom Sawyer tbh. Let's just keep it at 16.

zstd/seqdec_copy_test.go
klauspost, line 1: We already have `zstd/seqdec_test.go` for tests there. Don't want a bunch of random feature test files.

zstd/seqdec_asm.go
```diff
@@ -83,8 +83,9 @@ const errorOverread = 6
 // When the output and literal buffers have compressedBlockOverAlloc (16)
 // bytes of slack past their logical use, the assembly may use extended
 // memory copies that read and write in 16-byte blocks, overrunning the end
-// of a literal run or match by up to 15 bytes. Otherwise it must use the
-// bounds-exact ("safe") copies. This mirrors the analogous, always-dynamic
+// of a literal run or match by up to 15 bytes; an empty literal run still
+// copies one block, reading and writing the full 16 bytes of slack past
+// its position. Otherwise it must use the bounds-exact ("safe") copies. This mirrors the analogous, always-dynamic
 // selection in executeSimple below.
```
klauspost:
```suggestion
// its position. Otherwise it must use the bounds-exact ("safe") copies.
// This mirrors the analogous, always-dynamic selection in executeSimple below.
```

klauspost (changes requested):
Needs a couple of hours of fuzz testing as well. Rest lgtm.

zstd/seqdec_copy_test.go
lizthegrey: Done in 35ab49f: the test now lives in seqdec_test.go and the separate file is gone. Also rebased onto master so it sits on top of #1205.

zstd/seqdec_asm.go
lizthegrey: Applied in 35ab49f.

lizthegrey:
> Needs a couple of hours of fuzz testing as well. Rest lgtm.

Kicking off 8 hours of fuzz testing each on an AMD 5600X (Zen 3), AMD 4750G (Zen 2), and Ampere Altra (N1 core), will report tonight.

lizthegrey:
Fuzzed this branch (35ab49f, rebased on #1205) on three hosts, ~24.5 host-hours / ~167 worker-hours, ~3.5 billion plain and ~33 million asan executions, no crashes and no new corpus files from the stock harnesses.

**amd64, Ryzen 7 PRO 4750G (Zen 2), 12 workers, 8 h 40 min:** FuzzDecodeAll, FuzzDecAllNoBMI2, FuzzDecoder, FuzzNoBMI2Dec at 100 min each (418 M, 410 M, 410 M, 412 M execs), then FuzzDecodeAll and FuzzDecAllNoBMI2 under -asan for 60 min each (12.3 M, 12.9 M).

**amd64, Ryzen 5 5600X (Zen 3), 4 workers, 3 h 12 min:** FuzzDecodeAll 90 min (218 M) clean; FuzzDecAllNoBMI2 for 43 + 58 min (107 M + 155 M) across two attempts that host reboots cut short, no failures.

**arm64, Neoverse N1, 4 workers, 6 h:** FuzzDecodeAll and FuzzDecoder at 120 min each (226 M, 222 M), then both under -asan for 60 min each (4.0 M, 3.9 M).

**Extra differential harnesses (Zen 2, 4 workers, 6 h 40 min):** one forcing the extended-copy DecodeAll path with a pre-allocated dst, 240 min, 269 M, clean; one forcing the fully-async streaming path against a sync decode, 160 min, ~190 M, which surfaced #1206 — pre-existing on master and unrelated to this change, fix in #1207. Seed-corpus path counts show the stock harnesses reach the extended-copy assembly on 52% of blocks via DecodeAll and 80% via the streaming decoder, the rest on the bounds-exact assembly and the Go loop.

klauspost (approved):

PR #1212, opened 2026-09-10
zstd: two-pass decode with match-source prefetch on arm64 (largest win decoding SpeedBestCompression-class data)

lizthegrey:
### What

The synchronous decoder (`DecodeAll`, `Reader` with concurrency 1) runs one loop that decodes and executes each sequence. The concurrent decoder's two-pass shape (decode into `seqVals`, then `executeSimple`) knows every match offset before the copies run, so its execute loop can prefetch match sources and keep several misses in flight. This PR lets the synchronous decoder use that shape when it pays. **It is a general arm64 optimization, not an N1-specific one** — see Results below; an earlier revision of this PR carried a Neoverse-N1-only threshold override read from sysfs, which is gone (no CPU-model-specific code remains).

- `executeSimple` prefetches the match source eight sequences ahead (the knee of a distance sweep on a Neoverse N1) and both cache lines of it — tested against one line: on real text the two are a wash, but on far-offset blocks one line alone costs +17.5%, since the source often straddles a line boundary and the second line's miss no longer overlaps the first's. A per-call flag folds into the loop's existing bound check, so with the prefetch off no instruction beyond that check runs. New state lives in stack slots: the loop already uses the whole register file and avo cannot spill.
- A match reaching into a separate history buffer is redirected there rather than prefetched below the output buffer, where each such prefetch cost a page walk.
- The block takes the two-pass path when the window is at least 1 MiB **and** the block's offset table puts at least 18% of its sequences past 1 MiB (offset code 20). The offset FSE table's normalized counts are the encoder's histogram of the block's offset codes, so this is a cheap loop over the table (skipped outright when the table's symbol range can't reach the far code at all); predefined and RLE tables carry no histogram and fall back to the window rule. Same shape as C's `ZSTD_getOffsetInfo` selection.
- The avo fork gains the `PREFETCHT0` to `PRFM` lowering (honeycombio/avo 6461c92).

### Why the numbers below look modest on real text, and when they don't

`SpeedDefault`'s encoder (`doubleFastEncoder`) evicts hash-table entries fast enough that a match surviving to 1 MiB back is found only rarely — an offset histogram over real silesia text at `SpeedDefault` puts ≥1 MiB offsets at ~2.4% of sequences. That's why the gate opens on few blocks of most files below, and the aggregate win on real `SpeedDefault` text is close to noise: the code path this PR speeds up mostly isn't reached at that compression level. It's not dormant, though — recompressing the same files at `SpeedBestCompression` (same 8 MiB window) lets the encoder actually find those matches, and the gate opens on most of some files' blocks (webster 94.7%, dickens 76.2%, mr 58.7%, samba 18.7%, mozilla 7.9%, others 0% — genuinely data-dependent, not close to a coin flip) with a large, consistent win where it does:

| arm64 part (real hardware) | silesia geomean, SpeedBestCompression | dickens | webster | mr |
|---|---|---|---|---|
| Ampere Altra, N1 | −7.7% | −28.1% | −33.8% | −13.3% |
| AWS Graviton2, N1 | −5.0% | −20.8% | −26.9% | −9.3% |
| AWS Graviton3, V1 | −4.8% | −17.5% | −19.6% | −5.9% |

Confirmed on three separate physical cores across two vendors, including a non-N1 core (Graviton3/V1) — this is a general win when the gate actually fires, not an artifact of the box this was developed on. The block-level decision itself is identical across all three (it's driven by the bitstream's offset table, not the hardware). What's still open: how much real klauspost/compress traffic runs above `SpeedDefault` — that determines how often this matters in practice, and isn't something measurable from here.

### Results: `DecodeAll`, checksum off, `-count 5`, versus one-pass, SpeedDefault

| arm64 part | silesia subset¹ | silesia geomean² | offsets 1–2 MiB | offsets 128–768 KiB | `benchdecoder` corpus |
|---|---|---|---|---|---|
| Ampere Altra, N1 | ~0% | −0.3% | −39.7% | ~0% | −1.2% |
| AWS Graviton2, N1³ | — | — | — | — | −1.1% |
| Graviton3, V1 | ~0% | −0.1% | −31.7% | −1.0% | −0.7% |
| Graviton4, V2 | 0.0% | +0.3% | −24.1% | +1.1%⁴ | −0.8% |
| Graviton5, V3 | +0.6% | −0.5% | −17.5% | +3.2%⁴ | −0.6% |

¹ dickens+reymont+xml+ooffice+webster at SpeedDefault, 8 MiB window. ² 12 files, geomean within noise on every part at this threshold — see the SpeedBestCompression numbers above for where the gate actually opens on real text. ³ not re-measured on this exact corpus after the N1-specific threshold was removed; Graviton2 numbers above (SpeedBestCompression) and Ampere Altra's SpeedDefault row are the evidence this isn't N1-specific and isn't N1-dependent. ⁴ no block selected: run-to-run drift on a cloud box.

The 128–768 KiB column used to carry a Neoverse-N1-only win (−26.8% on the Altra) from a lower per-core threshold read from sysfs; that's gone; N1 now uses the same 1 MiB threshold as every other core and keeps the 1–2 MiB win, which is the bulk of what this PR targets.

| amd64 part, two-pass forced | silesia geomean | offsets 1–2 MiB |
|---|---|---|
| Sapphire Rapids (c7i) | +11.7% | −12.5% |
| Granite Rapids (c8i) | +10.3% | −8.8% |
| Zen 4 (m7a) | +15.5% | +9.4% |

Every silesia file loses on all three, so no gate helps there; the disabled prefetch check in the concurrent decoder's execute loop measures as noise (+2.4%, −3.1%, −0.2%).

Re-tested at the same `SpeedBestCompression` level that makes the arm64 numbers above real, not just `SpeedDefault` (Sapphire Rapids, real per-block gate, not forced): still no effect worth gating on. Forced two-pass on the `SpeedBestCompression` corpus is close to neutral overall (+0.23% geomean) but stays a genuine mixed bag per file — dickens −12.5% and webster −12.1% win, mozilla/nci/ooffice/samba/xml lose 6.5–9.4%. Swept the offset-share threshold from 18% (arm64's own value) to 90% looking for a cutoff where it nets a clean win: none exists, every value sits in a ±0–2% band with no monotonic trend — raising the bar prunes dickens/webster's own marginal blocks as fast as it excludes everyone else's losses. x86's wider out-of-order core hides most of the miss latency the one-pass loop would otherwise expose, so the two-pass restructuring's fixed overhead and its benefit are too close in magnitude for any share cutoff to separate cleanly. amd64 stays off — confirmed at both compression levels, not an artifact of only having tested `SpeedDefault`.

### Ruled out

C's `ZSTD_decompressSequencesLong` shape, a prefetch ring inside the one-pass loop (decode eight ahead, `ZSTD_prefetchMatch`, execute eight behind), was built for `decodeSync`, made error-order exact, and measured on the N1: it loses to the two-pass path on every input (silesia geomean +3% to +6% against one-pass, +12–15% on small windows). Its per-sequence bookkeeping costs about 10 cycles in a 60-cycle iteration, and the short two-pass execute loop overlaps misses on its own. Not in this PR.

### Tests

- `TestDecodeTwoPassMatchesOnePass`: both paths forced on any architecture through every `DecodeAll` and `Reader` shape, including a separate history buffer, plus corrupt and truncated input.
- `TestSequenceDecsUseTwoPass`: the decision for each offset-table kind against a small and a large window, including the no-far-codes-in-table quick reject.
- `Test_seqdec_execute`: prefetch off and on, output compared byte for byte (exercises the history redirect).
- `FuzzDecodeAll` with two-pass forced, 10 minutes, clean. Short suite, `-race` and `-tags noasm` on amd64; short suite on Ampere Altra and real AWS Graviton2/Graviton3.

## Summary by CodeRabbit

- **Performance**
  - Improved decompression performance for large-window data on supported hardware.
  - Added adaptive decoding and memory prefetching to accelerate processing of long-distance matches.
  - Preserved an optimized fallback path for platforms without specialized support.

- **Reliability**
  - Expanded validation to confirm consistent output across decoding strategies, buffer configurations, concurrency levels, and malformed or truncated input.

### DIFF
--- a/zstd/_generate/gen.go
+++ b/zstd/_generate/gen.go
@@ -3,7 +3,6 @@ package main
 //go:generate go run gen.go -out ../seqdec.s -arch amd64,arm64 -pkg=zstd
 
 import (
-	"flag"
 	"fmt"
 	"runtime"
 
@@ -48,9 +47,13 @@ const compressedBlockOverAlloc = 16
 // size of struct seqVals
 const seqValsSize = 24
 
-func main() {
-	flag.Parse()
+// prefetchDist is how many sequences ahead executeSimple prefetches the
+// match source: the knee of a distance sweep on a Neoverse N1 (see
+// executeSimple.generateProcedure). Whether the prefetch runs is decided
+// per call by executeAsmContext.prefetch.
+const prefetchDist = 8
 
+func main() {
 	Constraint(buildtags.Not("appengine").ToConstraint())
 	Constraint(buildtags.Not("noasm").ToConstraint())
 	Constraint(buildtags.Term("gc").ToConstraint())
@@ -1036,6 +1039,41 @@ func (e executeSimple) generateProcedure(name string) {
 	histBase := GP64()
 	histLen := GP64()
 
+	// Match-source prefetch. seqVals already holds every sequence of the
+	// block, so a running sum of ll+ml over the next prefetchDist entries
+	// names the output position each future match is copied to, and that
+	// position minus mo names its source. Touch the source (both halves of
+	// a possibly line-straddling copy) prefetchDist sequences before the
+	// copy needs it.
+	//
+	// The helper costs about a quarter of an iteration on data with nothing
+	// to hide (small windows), so the caller switches it per call through
+	// ctx.prefetch, which folds into the bound check below: with the
+	// prefetch off, the end pointer is the start pointer and every helper
+	// takes its skip branch.
+	//
+	// Register budget: the copy code already peaks at the full register
+	// file and avo has no spilling, so any new loop-carried register is a
+	// generate-time failure (seen with three, two and one). Nothing new is
+	// loop-carried in a register. The prefetch cursor always sits exactly
+	// 24*prefetchDist bytes ahead of seqsBase, so it is a displacement; the
+	// loop-invariant end pointer and the running output position live in
+	// stack slots, reloaded (and the latter stored back) by the helper, three
+	// L1 accesses per sequence.
+	//
+	// A match whose offset reaches past the start of out lives in the
+	// separate history buffer (streaming decode; DecodeAll keeps the whole
+	// frame in out and never takes this path). Its source is histEnd -
+	// (mo - pos), which is (q - mo) + (histEnd - outStart) for q the match's
+	// absolute output pointer: a fixed offset from the in-out address, so
+	// the helper redirects with one compare and one add. Left as garbage
+	// addresses below out, those prefetches cost a page walk each and made
+	// the small-block seqs.zip benchmark 70% slower.
+	seqsEndSlot := AllocLocal(8)
+	pfOutSlot := AllocLocal(8)
+	outStartSlot := AllocLocal(8)
+	histSkewSlot := AllocLocal(8)
+
 	{
 		ctx := Dereference(Param("ctx"))
 		tmp := GP64()
@@ -1043,6 +1081,19 @@ func (e executeSimple) generateProcedure(name string) {
 		TESTQ(seqsLen, seqsLen)
 		JZ(LabelRef("empty_seqs"))
 		Load(ctx.Field("seqs").Base(), seqsBase)
+		{
+			Comment("seqsEnd = seqsBase + 24 * len(seqs) if ctx.prefetch, else seqsBase (prefetch off)")
+			seqsEnd := GP64()
+			MOVQ(seqsBase, seqsEnd)
+			Load(ctx.Field("prefetch"), tmp)
+			TESTQ(tmp, tmp)
+			JZ(LabelRef("prefetch_off"))
+			LEAQ(Mem{Base: seqsLen, Index: seqsLen, Scale: 2}, seqsEnd) // * 3
+			SHLQ(U8(3), seqsEnd)                                        // * 8
+			ADDQ(seqsBase, seqsEnd)
+			Label("prefetch_off")
+			MOVQ(seqsEnd, seqsEndSlot)
+		}
 		Load(ctx.Field("seqIndex"), seqIndex)
 		Load(ctx.Field("out").Base(), outBase)
 		Load(ctx.Field("literals").Base(), literals)
@@ -1062,7 +1113,59 @@ func (e executeSimple) generateProcedure(name string) {
 		ADDQ(outPosition, outBase)
 	}
 
+	var prefetch func(n int)
+	{
+		MOVQ(outBase, pfOutSlot)
+		{
+			t := GP64()
+			MOVQ(outBase, t)
+			SUBQ(outPosition, t) // &out[0]
+			MOVQ(t, outStartSlot)
+			NEGQ(t)
+			ADDQ(histBase, t) // histEnd - &out[0]
+			MOVQ(t, histSkewSlot)
+		}
+		// prefetch touches the source of the sequence n entries past
+		// seqsBase, if there is one, and advances the pending output
+		// position past it.
+		prefetch = func(n int) {
+			skip := fmt.Sprintf("prefetch_skip_%d", n)
+			seq := Mem{Base: seqsBase, Disp: n * seqValsSize}
+			v := GP64()
+			t := GP64()
+			MOVQ(seqsEndSlot, v)
+			LEAQ(seq, t)
+			CMPQ(t, v)
+			JAE(LabelRef(skip))
+			inOut := fmt.Sprintf("prefetch_in_out_%d", n)
+			w := GP64()
+			MOVQ(pfOutSlot, t)
+			MOVQ(seq.Offset(0*8), v) // ll
+			ADDQ(v, t)               // t = the match's output pointer
+			MOVQ(t, w)
+			MOVQ(seq.Offset(2*8), v) // mo
+			SUBQ(v, w)               // w = its source, if inside out
+			MOVQ(outStartSlot, v)
+			CMPQ(w, v)
+			JAE(LabelRef(inOut))
+			MOVQ(histSkewSlot, v)
+			ADDQ(v, w) // redirect into the history buffer
+			Label(inOut)
+			PREFETCHT0(Mem{Base: w})
+			PREFETCHT0(Mem{Base: w, Disp: 64})
+			MOVQ(seq.Offset(1*8), v) // ml
+			ADDQ(v, t)               // past the match
+			MOVQ(t, pfOutSlot)
+			Label(skip)
+		}
+		Comment("Prime the prefetch window")
+		for n := 0; n < prefetchDist; n++ {
+			prefetch(n)
+		}
+	}
+
 	Label("main_loop")
+	prefetch(prefetchDist)
 
 	moPtr := Mem{Base: seqsBase, Disp: 2 * 8}
 	mlPtr := Mem{Base: seqsBase, Disp: 1 * 8}
--- a/zstd/seqdec.go
+++ b/zstd/seqdec.go
@@ -76,6 +76,20 @@ type sequenceDecs struct {
 	maxSyncLen   uint64
 }
 
+const twoPassFarCode = 17 // offsets of 128 KiB and up
+
+// useTwoPass reports whether the block is decoded in two passes with the
+// match-source prefetch (seqdec_arm64.go): a window of at least
+// decodeTwoPassMinWindow and, unless the offset table is predefined or
+// RLE, a far-code share of at least twoPassMinFarShare.
+func (s *sequenceDecs) useTwoPass() bool {
+	if s.windowSize < decodeTwoPassMinWindow {
+		return false
+	}
+	share := s.offsets.fse.codeShare(twoPassFarCode)
+	return share < 0 || share >= twoPassMinFarShare
+}
+
 // initialize all 3 decoders from the stream input.
 func (s *sequenceDecs) initialize(br *bitReader, hist *history, out []byte) error {
 	if err := s.litLengths.init(br); err != nil {
--- a/zstd/seqdec.go
+++ b/zstd/seqdec.go
@@ -76,12 +76,15 @@ type sequenceDecs struct {
 	maxSyncLen   uint64
 }
 
-const twoPassFarCode = 17 // offsets of 128 KiB and up
+// twoPassFarCode is the offset code from which a match counts as far: 20 is
+// 1 MiB, the distance every measured arm64 core gains from; the Neoverse N1
+// gains from 128 KiB (17) and is detected in seqdec_linux_arm64.go.
+var twoPassFarCode = 20
 
 // useTwoPass reports whether the block is decoded in two passes with the
 // match-source prefetch (seqdec_arm64.go): a window of at least
 // decodeTwoPassMinWindow and, unless the offset table is predefined or
-// RLE, a far-code share of at least twoPassMinFarShare.
+// RLE, a share of at least twoPassMinFarShare of codes twoPassFarCode+.
 func (s *sequenceDecs) useTwoPass() bool {
 	if s.windowSize < decodeTwoPassMinWindow {
 		return false
--- a/zstd/seqdec_linux_arm64.go
+++ b/zstd/seqdec_linux_arm64.go
@@ -1,0 +1,29 @@
+//go:build linux && arm64 && !appengine && !noasm && gc
+
+package zstd
+
+import (
+	"os"
+	"strconv"
+	"strings"
+)
+
+// The two-pass decode wins on real data only on the Neoverse N1 (Graviton2,
+// Ampere Altra), whose core exposes far-source misses from 128 KiB; the
+// wider V1, V2 and V3 cores hide those and lose 5-15% on the same data, and
+// gain only from 1 MiB. Linux exposes the core's MIDR register in sysfs.
+func init() {
+	if b, err := os.ReadFile("/sys/devices/system/cpu/cpu0/regs/identification/midr_el1"); err == nil && isNeoverseN1(string(b)) {
+		twoPassFarCode = 17
+	}
+}
+
+// isNeoverseN1 reports whether a MIDR_EL1 value names an Arm (0x41)
+// Neoverse N1 (part 0xd0c).
+func isNeoverseN1(midr string) bool {
+	v, err := strconv.ParseUint(strings.TrimPrefix(strings.TrimSpace(midr), "0x"), 16, 64)
+	if err != nil {
+		return false
+	}
+	return v>>24&0xff == 0x41 && v>>4&0xfff == 0xd0c
+}
--- a/zstd/_generate/go.mod
+++ b/zstd/_generate/go.mod
@@ -15,4 +15,6 @@ require (
 
 replace github.com/klauspost/compress => ../..
 
+// The fork carries the arm64 lowering printer (branch lizf.arm64-lowering-printer);
+// the replacement goes away once that is upstreamed.
 replace github.com/mmcloughlin/avo => github.com/honeycombio/avo v0.6.1-0.20260910220642-6461c9239401
--- a/zstd/_generate/go.mod
+++ b/zstd/_generate/go.mod
@@ -15,4 +15,4 @@ require (
 
 replace github.com/klauspost/compress => ../..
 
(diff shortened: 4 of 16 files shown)


lizthegrey:
> 🧹 Nitpick comments (1)
> > zstd/_generate/go.mod (1)> `18-18`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _💤 Low value_
> > > **Record the upstreaming target for the Avo fork.**
> > > The pseudo-version is fetchable. Add an issue link or tracking note for upstreaming the arm64 `PRFM` lowering so future regeneration can remove the `github.com/honeycombio/avo` replacement.

Yeah, yeah, we know, Coderabbit, stop nagging us about this one. Ongoing out-of-band discussions about how the avo fork will be maintained going forward.

zstd/_generate/gen.go
```diff
@@ -1036,13 +1039,61 @@ func (e executeSimple) generateProcedure(name string) {
...
+			MOVQ(seqsBase, seqsEnd)
+			Load(ctx.Field("prefetch"), tmp)
+			TESTQ(tmp, tmp)
+			JZ(LabelRef("prefetch_off"))
+			LEAQ(Mem{Base: seqsLen, Index: seqsLen, Scale: 2}, seqsEnd) // * 3
+			SHLQ(U8(3), seqsEnd)                                        // * 8
+			ADDQ(seqsBase, seqsEnd)
+			Label("prefetch_off")
+			MOVQ(seqsEnd, seqsEndSlot)
+		}
```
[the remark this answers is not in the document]
lizthegrey: Not out of frame. The Go arm64 assembler reserves the declared frame plus 8 bytes for the saved link register, padded to 16, and the lowering printer relies on that: its `frameAdjust` places avo's locals from 8(RSP) for exactly this reason. Assembled prologue of `sequenceDecs_executeSimple_arm64` from `go tool objdump`:

```
TEXT ·sequenceDecs_executeSimple_arm64(SB), $32-9
  MOVD.W R30, -48(RSP)   // 48-byte frame: LR at 0, locals 8..47
  ...
  MOVD R10, 32(RSP)      // histSkewSlot, bytes 32..39, inside the frame
```

The same layout has held for `decodeSync_arm64` (`$96`, slots to 96(RSP)) since the lowering landed, and the full suite plus the equivalence tests pass on five arm64 machines with this path exercised.

klauspost:
> Yeah, yeah, we know, Coderabbit, stop nagging us about this one. Ongoing out-of-band discussions about how the avo fork will be maintained going forward.

@lizthegrey What do you think of my suggestion to have a fork until it gets merged? You are welcome to replace imports in code, instead of having the go.mod replace(s)?

(if you @ the bot it will remember, like this...) 

@coderabbitai Thanks for bringing it up, for now we allow the replace for the generator, but I would like a more "official" source as well. But you can ignore this one replace for now.

lizthegrey:
> @lizthegrey What do you think of my suggestion to have a fork until it gets merged? You are welcome to replace imports in code, instead of having the go.mod replace(s)?

I told Michael about intentions to soft-fork soon if he is still too busy to maintain avo, so probably in another two weeks if there's no movement I will.

lizthegrey:
> I told Michael about intentions to soft-fork soon if he is still too busy to maintain avo, so probably in another two weeks if there's no movement I will.

Michael is now merging PRs, so I expect to not soft-fork.

zstd/seqdec.go
```diff
@@ -76,6 +76,23 @@ type sequenceDecs struct {
...
+// 1 MiB, the distance every measured arm64 core gains from; the Neoverse N1
+// gains from 128 KiB (17) and is detected in seqdec_linux_arm64.go.
+var twoPassFarCode = 20
+
+// useTwoPass reports whether the block is decoded in two passes with the
+// match-source prefetch (seqdec_arm64.go): a window of at least
+// decodeTwoPassMinWindow and, unless the offset table is predefined or
+// RLE, a share of at least twoPassMinFarShare of codes twoPassFarCode+.
+func (s *sequenceDecs) useTwoPass() bool {
+	if s.windowSize < decodeTwoPassMinWindow {
```
klauspost, line 91: I think a quick reject on `twoPassMinFarShare` will minimize the check time by avoiding finding the codeshare, if impossible.

zstd/seqdec_linux_arm64.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+// The two-pass decode wins on real data only on the Neoverse N1 (Graviton2,
+// Ampere Altra), whose core exposes far-source misses from 128 KiB; the
+// wider V1, V2 and V3 cores hide those and lose 5-15% on the same data, and
+// gain only from 1 MiB. Linux exposes the core's MIDR register in sysfs.
+func init() {
+	if b, err := os.ReadFile("/sys/devices/system/cpu/cpu0/regs/identification/midr_el1"); err == nil && isNeoverseN1(string(b)) {
+		twoPassFarCode = 17
+	}
```
klauspost: Not a fan of reading files on init. Should be lazy and limit the input size.

zstd/_generate/gen.go
```diff
@@ -1062,7 +1113,59 @@ func (e executeSimple) generateProcedure(name string) {
...
+			MOVQ(seq.Offset(2*8), v) // mo
+			SUBQ(v, w)               // w = its source, if inside out
+			MOVQ(outStartSlot, v)
+			CMPQ(w, v)
+			JAE(LabelRef(inOut))
+			MOVQ(histSkewSlot, v)
+			ADDQ(v, w) // redirect into the history buffer
+			Label(inOut)
+			PREFETCHT0(Mem{Base: w})
+			PREFETCHT0(Mem{Base: w, Disp: 64})
```
klauspost, line 1155: I get that it will try to have at least 2 cachelines available, but it will also fetch a lot of data that isn't needed. Was "2 cachelines" determined by testing?

klauspost (changes requested):
Thanks for the extensive testing.

Reading your numbers, it honestly looks to me like a failed experiment.

Optimizing specifically for a single arm64 CPU type doesn't IMO belong in this package.

You are welcome to keep working on it, but it seems like a "problem" that will phase itself out as these CPUs are sunsetted.

Added notes if you want to keep working on this.

lizthegrey:
> Optimizing specifically for a single arm64 CPU type doesn't IMO belong in this package.

Noted, I can remove the N1 exception and focus on the general case, although it *is* still roughly 40% of our AWS Lambda fleet.

> You are welcome to keep working on it, but it seems like a "problem" that will phase itself out as these CPUs are sunsetted.
> 
> Added notes if you want to keep working on this.

The rest of the optimisation is still good even with the N1 special threshold/kludge removed, I believe.

klauspost:
@lizthegrey Maybe I misread your PR text then. Seemed it was only really a benefit on N1. If you could have your LLM redo the text, maybe that would clear things up. But yes, no CPU-model specific code, please.

lizthegrey:
> @lizthegrey Maybe I misread your PR text then. Seemed it was only really a benefit on N1. If you could have your LLM redo the text, maybe that would clear things up. But yes, no CPU-model specific code, please.

Yeah, there's little to no improvement on silesia as a whole *at default compression settings*, but the microbenchmark does show that far offsets improve dramatically with the prefetch. I'll go and pull some updated numbers. clearly this does come up in practice or else Yann wouldn't have implemented it in the C code.

edit: found the problem, you need to compress silesia with *best* compression settings, then the change in decompression speed is very much visible.

zstd/seqdec.go
lizthegrey: Added — `useTwoPass` now checks the offset table's `symbolLen` against `twoPassFarCode` before calling `codeShare`: if no symbol in the table reaches the far code, `codeShare` could only return 0, so the scan is skipped entirely. Pushed as 0104600.

zstd/seqdec_linux_arm64.go
lizthegrey: Removed — `twoPassFarCode` is back to a plain `const` at 20 (1 MiB), no sysfs read, no per-core branch. Pushed as 0104600.

Quantified what N1 gives up: on oracle (Ampere Altra, an N1), real silesia data at farCode=20 is noise-level versus one-pass (geomean −0.33%, vs. the −3.8% the 17-threshold gave up). Breaking that down by offset range: the synthetic 128–768 KiB bucket the N1-specific threshold targeted goes from −26.8% down to ~0% (that gain is fully given up), while the 1–2 MiB bucket both thresholds already qualify for stays at −39.7% (kept, since it clears 1 MiB either way). N1 keeps the general-arm64 win this PR is actually about and only gives up the narrow sub-1MiB extra that needed the CPU-specific override.

zstd/_generate/gen.go
lizthegrey: It wasn't — the D-sweep tested prefetch *distance*, not line count, and the second line was a design assumption ("matches straddle cache lines"), not a measured one. Tested it now on oracle: 1 line vs. 2 on real silesia data is a wash (geomean −0.39%, i.e. slightly *faster* with 1 line — the second line is often wasted fetch). But on the synthetic 1–2 MiB offset bucket — the actual target of this gate — dropping to 1 line costs +17.5%: those matches are long/far enough that the source often does straddle a line boundary, and missing the second line serializes what should be a hidden DRAM-class miss instead of prefetching both lines in parallel. Keeping 2 lines.

lizthegrey:
Fair to push on this, so I checked how much of the gate's value was actually N1-specific versus real for the general case. Rewrote the PR description with the details; summary here.

The CPU-specific sysfs override is gone — `twoPassFarCode` is a plain `const` at 20 (1 MiB) for every arm64 core now, no per-core branch.

On why the numbers looked N1-only: this PR's own benchmarks are all `SpeedDefault`, where the encoder structurally can't find most far-offset matches (hash-table eviction means a match surviving to 1 MiB back happens on roughly 2.4% of sequences on real text). So on real silesia text at `SpeedDefault` the two-pass gate mostly stays dormant and the aggregate win is close to noise — that part of your read was correct, and it's not an N1 vs. non-N1 thing, it's a "this code path is rarely reached at this compression level" thing.

Where it isn't dormant: recompressed the same silesia files at `SpeedBestCompression` (same 8 MiB window), where the encoder can actually reach those offsets on ordinary text. Several files then take the two-pass path on most of their blocks (webster 94.7%, dickens 76.2%, mr 58.7% — genuinely data-dependent, 0% on the rest) and the win is real: geomean −7.7% on oracle (Ampere Altra, N1), −5.0% on a real AWS Graviton2 (N1), and −4.8% on a real AWS Graviton3 (V1 — not an N1-class core at all). Same direction, same files, comparable magnitude across two vendors and two different core families. The block-level gate decision itself is identical everywhere, since it's driven by the bitstream's offset table, not the hardware.

So: real, reproducible, general-arm64 benefit when the workload reaches it, rare at `SpeedDefault`, common at higher compression levels. What I can't answer from here is how much real traffic runs above `SpeedDefault` — flagged as open in the PR description rather than asserting either way. Happy to keep iterating with the CPU-specific case gone.

zstd/_generate/gen.go
klauspost: I am not interested in optimizing towards artificial targets..

klauspost:
Sure. I'm mostly interested in real-world scenarios. Meaning full streams or real data. 

Not a particular fan of silesia, but for this I'd expect it to be reasonable. But both that and enwik has a lot of short matches.

Please instruct your LLM to stop using artificially generated tests. You can "prove" pretty much anything with artificial tests.

You can download additional test sets: https://klauspost.com/files/compress/

Log data: https://klauspost.com/files/compress/cockroach.node1.log.zst https://klauspost.com/files/compress/apache.log.zst
Serialized: https://klauspost.com/files/compress/github-ranks-backup.bin.zst https://klauspost.com/files/compress/gob-stream.zst
JSON: https://klauspost.com/files/compress/github-june-2days-2019.json.zst
CSV: https://klauspost.com/files/compress/nyc-taxi-data-10M.csv.zst https://klauspost.com/files/compress/sofia-air-quality-dataset.tar.zst
DB dump: https://klauspost.com/files/compress/consensus.db.10gb.zst
VM Image: https://klauspost.com/files/compress/rawstudio-mint14.tar.zst
Mixed: https://klauspost.com/files/compress/10gb.tar.zst 

Most will have been compressed with `-19`, so they should have long offsets. You can of course recompress as you see fit.

There are some 7z files with individual smaller files.

Use what you need, but benchmark against *real* data.

lizthegrey:
Ran the corpora you linked, on real data end to end (no silesia/enwik, no synthetic generator). Two separate tests, both on `oracle` (Neoverse N1, same box as my earlier numbers):

**1. Your files, decoded as provided** (`.zst` as downloaded, whatever offsets/window your compressor produced — full size, not truncated). `go test -bench BenchmarkDecoderWithCustomFiles`, comparing this branch against `de8f55d` (tip before this PR):

| file | decompressed | geomean over 5 decode modes |
|---|---|---|
| apache.log | 2.6 GB | +4.1% |
| cockroach.node1.log | 10.5 GB | -0.1% |
| github-june-2days-2019.json | 6.3 GB | +3.5% |
| github-ranks-backup.bin | 1.9 GB | +26.8% |
| gob-stream | 1.9 GB | +11.5% |
| nyc-taxi-data-10M.csv | 3.3 GB | +69.5% |
| rawstudio-mint14.tar | 8.6 GB | +16.7% |
| sofia-air-quality-dataset.tar | 15.5 GB | +13.4% |

Geomean across all 40 (file × decode-mode) points: **+16.6%**. On just the single-threaded decode path (closest to where the two-pass gate actually fires): **+20-22%**. `cockroach.node1.log` is flat — makes sense, log lines repeat at short range. `nyc-taxi` and `github-ranks-backup` are where it's dramatic — long-range redundancy in columnar/tabular data.

**2. Isolating default vs. best-compression, with our own encoder.** This directly checks the thing you flagged: does `SpeedDefault` ever reach these offsets on real data, or is it a `SpeedBestCompression`-only effect? Recompressed a 512 MiB prefix of each of the same 8 corpora at both levels (capped so `SpeedBestCompression`'s slower encode finishes in reasonable time), decoded both ways, same branch-vs-`de8f55d` comparison, single-threaded path:

- **SpeedDefault, 8 files: geomean -0.2%** — noise, no measurable effect. Confirms your read: at default settings our encoder essentially never produces the far-offset share the gate needs.
- **SpeedBestCompression, 8 files: geomean +10.7%** — gob-stream +21.5%, nyc-taxi +73.2%, sofia +7.8%, rawstudio +3.3%; three files (cockroach, github-ranks-backup, json) land at -1 to -2%, which I'd call within noise for a single `benchtime=1x` sample rather than a real regression.

So: real, reproducible, general-arm64 (not N1-specific — the CPU-model branch is already gone from this PR), gated behind `SpeedBestCompression`-class offset distributions, and now shown across 8 independent real-world corpora instead of just silesia. Still open: how much real decode traffic runs above `SpeedDefault` — I can't answer that from here.

This was a single sample per point (`-benchtime=1x`), which is enough to see the shape but not to defend the noise-level deltas. Happy to follow up with a `-count=10` + `benchstat` pass on whichever of these you want higher confidence on.

klauspost:
Are the percentages time or throughput?

lizthegrey:
> Are the percentages time or throughput?

throughput. walltime can be derived from it though (`1/(1-n)`) etc.

lizthegrey:
Yes. On real production data, this PR makes decode 4–15% faster.

Setup: our production string columns (log/trace data), on Graviton5 (`0xd84`), decoded through our production `DecodeAll` per-frame path. Throughput in decompressed MiB/s, this branch against `de8f55d`, two alternating runs each.

| level, frame = window | this PR | `de8f55d` | change |
|---|---|---|---|
| Better, 1 MiB frame, 64 KiB window (control) | ~1150 | ~1150 | 0% |
| Better, 1 MiB | 1242 | 1108 | +12% |
| Better, 8 MiB | 2027 | 1952 | +4% |
| Best, 8 MiB | 1955 | 1704 | +15% |
| Better, 32 MiB | 1875 | 1932 | −3% |

This node was also serving production traffic, and repeat runs of the same binary varied by ±10%, so the +4% and −3% rows are within noise. The gain tracks how many far matches the data has. That matters for us because we're moving from a 64 KiB to an 8 MiB window, which alone cuts these files ~48%.

Co-authored-by: Claude Opus 5.5 <noreply@anthropic.com>

klauspost (approved):

PR #1213, opened 2026-09-11
s2: Swap appendTo() argument order to match add()

Jille:
I chose add()'s order as the winning order because it aligns with the return values of Find().

## Summary by CodeRabbit

- **Refactor**
  - Switched index metadata serialization so compressed and uncompressed size totals are in same order as add method.


klauspost:
@coderabbitai Why are you classifying this as bug fixes?

@Jille I suggest you find something more substantial. This is similar to sending a comma fix to documentation. While technically correct, you are taking up a lot of time with little benefit for the project.

Jille:
The context is that I'd extracted your index code to build my own S2 compatible streaming format to help us switch from S2 to minlz and zstd. I had to expose add() and appendTo() and made a bug because of the argument order.
This change is to prevent anyone in the future from having that problem. (Or worse: if you'd decide to expose the methods and then be locked in to the inconsistency forever.)

But I understand that from your point of view it's low value.

PR #1223, opened 2026-09-17
gzhttp: only decompress a request when gzip is the outermost coding

youdie006:
`contentGzip` (`gzhttp/compress.go:992`) decides whether to gunzip a request body with `parseEncodingGzip`, which finds `gzip` anywhere in a comma list and reads `;q=` weights. That is an Accept-Encoding parser, and `acceptsGzip` is right to use it. Content-Encoding lists codings in the order they were applied, so only the last one is the outermost coding a decoder may remove.

Measured through `NewWrapper(AllowCompressedRequests(true))`, what the wrapped handler sees today:

| sent | handler sees | body | read error |
|---|---|---|---|
| `deflate, gzip` | `""` | still deflate-compressed | none |
| `gzip, deflate` | `""` | empty | `gzip: invalid header` |
| `gzip` | `""` | decoded | none |

The second row fails the request. The first is quieter and worse: gunzipping was correct, but deleting the whole header removed the only signal that `deflate` is still applied, so the handler gets bytes it cannot interpret and no error.

This decides on the last coding and strips only that one, via the repo's own `parseCoding` so the accepted spellings for that element are unchanged. The single `Content-Encoding: gzip` case is byte-identical to before.

<details>
<summary>Verification</summary>

Base `de8f55df`. New test `gzhttp/compress_ce_test.go` drives the real wrapped handler rather than calling `contentGzip` directly.

| row | `gzhttp/compress.go` md5 | result |
|---|---|---|
| pristine | `0e8bd3a5b6e6ec74141166f903dc26f8` | FAIL |
| this PR | `e3a7e4ec681c0f6c4c77a693208860e7` | ok |
| detect with `parseEncodingGzip` again, keep the header rewrite | `5970a7ac11...` | FAIL |
| keep the order check, delete the whole header | `cf6108d168...` | FAIL |
| `IndexByte` instead of `LastIndexByte` | `bed1792c92...` | FAIL |

The last row needs the three-coding case to die; with only two codings the first and last comma agree and it survives, so `br, deflate, gzip` is in the table for that reason.

`go test -count=1 ./gzhttp/...` ok for `gzhttp`, `gzhttp/writer/gzkp`, `gzhttp/writer/gzstd`. `go vet ./gzhttp/` clean, `gofmt -l gzhttp/` empty. I did not run the repo-wide suite or the fuzz jobs; the fuzz workflow targets `s2` and `zstd` only and does not reach `gzhttp`.

One behaviour change on malformed input worth naming: a trailing comma such as `Content-Encoding: gzip,` used to decode and now does not, because the last element is empty.

</details>

Written with AI assistance (Claude); the measurements and the mutation rows above were run locally and I have reviewed the change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of compressed requests with multiple or repeated `Content-Encoding` header values.
  * The outermost compression layer is decoded correctly while inner encodings remain available in the request metadata.
  * Added coverage for combined, repeated, split-line, and multi-coding headers.


gzhttp/compress_ce_test.go
klauspost, line 1: Don't add new files.

klauspost (changes requested):
Your title and description indicates *decompression*, but I only see changes to *compression*. Are you leaving out some part, or do they work? Add roundtrip tests, not just sparse unit tests.

klauspost (approved):

PR #1224, opened 2026-09-18
flate: avoid FMA in EstimatedBits for portable rounding

### LINKED ISSUE #72: On some rare inputs, the output depends on the CPU architecture (x86_64 vs aarch64)
I was expecting pgzip to produce identical output given identical input (and identical compression settings).
I found that for a very small subset of inputs, the output differs between runs on x86_64 and aarch64.
I have attached a minimal reproduction.

I have since noticed that this also affects `klauspost/flate` directly, as well as the Go stdlib flate (since Go 1.27).

https://github.com/malt3/pgzip-cpu-dependent-output-repro

I believe the root cause has to do with this part of the Go spec:

> Floating-point operators[¶](https://go.dev/ref/spec#Floating_point_operators)
> 
> For floating-point and complex numbers, +x is the same as x, while -x is the negation of x. The result of a floating-point or complex division by zero is not specified beyond the IEEE 754 standard; whether a [run-time panic](https://go.dev/ref/spec#Run_time_panics) occurs is implementation-specific.
> 
> An implementation may combine multiple floating-point operations into a single fused operation, possibly across statements, and produce a result that differs from the value obtained by executing and rounding the instructions individually. An explicit [floating-point type](https://go.dev/ref/spec#Numeric_types) [conversion](https://go.dev/ref/spec#Conversions) rounds to the precision of the target type, preventing fusion that would discard that rounding.
> 
> For instance, some architectures provide a "fused multiply and add" (FMA) instruction that computes x*y + z without rounding the intermediate result x*y. These examples show when a Go implementation can use that instruction:

```
// FMA allowed for computing r, because x*y is not explicitly rounded:
r  = x*y + z
r  = z;   r += x*y
t  = x*y; r = t + z
*p = x*y; r = *p + z
r  = x*y + float64(z)

// FMA disallowed for computing r, because it would omit rounding of x*y:
r  = float64(x*y) + z
r  = z; r += float64(x*y)
t  = float64(x*y); r = t + z
```

[source](https://go.dev/ref/spec#Floating_point_operators)

malt3:
EstimatedBits and mFastLog2 had chained multiply-add expressions that the compiler is allowed to contract into fused multiply-adds.
FMA contraction rounds differently than sequential multiply+add, so EstimatedBits previously returned a different result on architectures with hardware FMA (e.g. arm64) than on default amd64, making EstimatedBits, and the resulting compressed output, GOARCH-dependent.

Fixes https://github.com/klauspost/pgzip/issues/72

## Summary by CodeRabbit

- **Tests**
  - Added an architecture-specific regression test to verify consistent floating-point results for compression calculations.

- **Refactor**
  - Reorganized an internal logarithm approximation without changing its arithmetic results or user-visible behavior.


flate/mfastlog2_arch_test.go
klauspost, line 1: No needless new file. Add to existing.

klauspost (commented):
What a mess.

malt3:
Let's hold off merging for a moment. I noticed that `mFastLog2` does indeed produce different output on different CPU architectures, but fixing this alone isn't enough to fix the issue in the reproducer (`pgzip` produces different output based on the testdata).
I'll investigate more.

klauspost:
Sure, no rush.

malt3:
@klauspost I found more places that are affected by FMA contraction.
I also updated the regression test to include the original test data and perform end-to-end flate compression.

klauspost:
Cool. I will check it out Monday.

klauspost (approved):

PR #1225, opened 2026-09-18
gzhttp: preserve caller request headers in Transport

jakezwang:
Clone the request before `gzhttp.Transport` adds its automatic `Accept-Encoding` header. The current transport modifies the caller's header map, contrary to the `http.RoundTripper` contract. When the same request is sent again after closing the response body, that injected header is mistaken for a caller-supplied preference and the second response is returned compressed.

Add a regression that reuses a request for gzip and zstd responses, checks both decoded bodies and `Uncompressed`, and verifies the original request headers remain unchanged. Both cases fail before the fix.

Validation: full tests and vet on Go 1.27.1; `gzhttp/...` race tests on Go 1.25.0 and Go 1.26.4; `gzhttp/...` tests with `noasm,nounsafe` on Go 1.27.1; Go 1.26.4 `go fix -diff ./...` and formatting checks are clean. Full `go test -cpu=1,4 -short -race -tags=nounsafe ./...` also passes on Go 1.27.1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Requests can now be safely reused across multiple attempts without modifying their original headers.
  - Compression negotiation and decompressed response handling remain unchanged for supported gzip and zstd responses.

- **Tests**
  - Added coverage confirming request reuse, response decompression, and preservation of custom request headers.


klauspost (approved):
lgtm

